PR #3, opened 2026-04-17
fix(db): drop FKs to removed users table so db:migrate completes

gomessguii:
## Summary

- Seven migrations still referenced a `users` table that no longer exists in community (removed in `9944be0` when identity moved to `evo-auth-service`). `db:migrate` aborted on the first of them (`CreateFacebookCommentModerations`), blocking `MakeAttachmentPolymorphic` and leaving `attachments` stuck with the legacy `message_id` column. Sidekiq/Puma then crashed with `PG::UndefinedColumn: attachments.attachable_id does not exist` on any media webhook or contact deletion.
- Each `foreign_key: { to_table: :users }` / `add_foreign_key …, :users` swapped for a plain `uuid` column (same name, so models keep working) plus manual indexes where the auto-generated ones went away.
- `docker/docker-entrypoint.sh`: collapsed the `db:create` / `db:migrate` / `db:evolution_prepare` trio (all masked by `2>/dev/null || true`) into a single unmasked `bundle exec rails db:evolution_prepare`. The mask is what hid the abort in the first place — a broken migration must stop the boot.

### Root cause

The account/user removal refactor (`9944be0`) only touched `db/schema.rb`; the downstream migrations that reference `:users` were never scrubbed. The `version: 9025_08_19_224901` bump in that same commit papers over the problem via `assume_migrated_upto_version` — it only works on the `db:schema:load` path. Any `db:migrate` on an empty DB still walks the migrations in chronological order and hits the FK abort.

## Test plan

- [x] `db:migrate` against a clean Postgres completes: 51/51 migrations applied, 0 pending.
- [x] `\d attachments` on the resulting DB shows `attachable_type` / `attachable_id` (with `index_attachments_on_attachable_type_and_attachable_id`) and no `message_id`.
- [x] `schema_migrations` contains `20251119155458` (the previously-blocked polymorphic migration).
- [ ] Deploy against an existing installation that already patched the columns manually: confirm `MakeAttachmentPolymorphic` idempotency (may need a follow-up guard).
- [ ] Smoke the Sidekiq webhook receiving media and the contact deletion flow end-to-end once the image is rebuilt.

### DIFF
--- a/db/migrate/20251114140050_create_facebook_comment_moderations.rb
+++ b/db/migrate/20251114140050_create_facebook_comment_moderations.rb
@@ -11,7 +11,7 @@ class CreateFacebookCommentModerations < ActiveRecord::Migration[7.1]
       t.string :action_type, null: false # 'delete_comment', 'block_user', 'send_response'
       t.text :response_content # Generated response content (if applicable)
       t.text :rejection_reason # Reason for rejection
-      t.references :moderated_by, null: true, foreign_key: { to_table: :users }, type: :uuid
+      t.uuid :moderated_by_id
       t.datetime :moderated_at
 
       t.timestamps
@@ -21,6 +21,6 @@ class CreateFacebookCommentModerations < ActiveRecord::Migration[7.1]
     add_index :facebook_comment_moderations, :moderation_type unless index_exists?(:facebook_comment_moderations, :moderation_type)
     add_index :facebook_comment_moderations, :comment_id unless index_exists?(:facebook_comment_moderations, :comment_id)
     add_index :facebook_comment_moderations, [:status, :moderation_type] unless index_exists?(:facebook_comment_moderations, [:status, :moderation_type])
-    # Note: conversation_id, message_id, and moderated_by_id indexes are automatically created by t.references
+    add_index :facebook_comment_moderations, :moderated_by_id unless index_exists?(:facebook_comment_moderations, :moderated_by_id)
   end
 end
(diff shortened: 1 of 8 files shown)


DavidsonGomes (approved):
LGTM. The diagnosis is clean: `9944be0` removed the `users` table from `db/schema.rb` and bumped `assume_migrated_upto_version`, but the migrations downstream still carried `foreign_key: { to_table: :users }` references. The schema-load path is fine (assume_migrated_upto_version masks it); the migrate-from-empty path is what breaks, exactly the path Sidekiq + a fresh community install hits. Aborts on `CreateFacebookCommentModerations` propagate forward and `MakeAttachmentPolymorphic` never runs — which is what produces the EVO-973 symptom (`PG::UndefinedColumn: attachments.attachable_id does not exist`).

Mechanical fix is correct: each `references :x, foreign_key: { to_table: :users }` swapped for `uuid :x_id` with manual indexes where the auto-generated FK index goes away. Same column names so AR models keep working; you lose DB-level cascade for nullify/destroy on user deletion, but in this architecture user lifecycle lives in evo-auth so the FK was symbolic anyway.

The docker-entrypoint cleanup is the right call too. Hiding `db:migrate` behind `2>/dev/null || true` is what kept this from being noticed for weeks. `db:evolution_prepare` is the single seam now and a broken migration will fail loudly instead of leaving the DB half-migrated.

**One thing to verify post-merge** (already on the test plan as unchecked) — `MakeAttachmentPolymorphic` idempotency on installs that manually patched columns. If anyone ran `ALTER TABLE attachments ADD COLUMN attachable_id` by hand, the migration must be idempotent or it'll abort on those installs. Worth a separate follow-up rather than blocking this PR — the migration itself isn't being modified here, only unblocked.

**Resolves EVO-973** (currently P1 Urgent, Todo, Guilherme).

PR #9, opened 2026-04-22
fix(crm): accept filters payload alias and add pipeline_id + contact_id (EVO-974)

nickoliveira23:
## Summary
- Conversation filter endpoint (`POST /conversations/filter`) was unusable for multi-row filters because the service iterated `@params[:payload]` while the evo frontend sends the rows under `:filters`. A `filter_payload` helper now accepts either key with a nil-safe guard — malformed bodies produce an empty relation instead of `NoMethodError: undefined method 'each' for nil`.
- Rewrote `FilterService#query_builder` so that trailing AND/OR emitted by per-attribute handlers is stripped, and the connector is re-injected as an infix between clauses. Fixes the pre-existing `cond_0  cond_1 AND` → invalid-SQL pattern for any multi-row filter.
- Registered `pipeline_id` in `filter_keys.yml` and added a dedicated `pipeline_filter_query` handler (EXISTS-checks `pipeline_items`). Previously `pipeline_id` raised `InvalidAttribute`.
- Registered `contact_id` in `filter_keys.yml` for the new Contact filter on the frontend (served by the generic `default_filter` handler).

## Security
- No new SQL string interpolation of user-controlled values. The `pipeline_filter_query` uses bind variables (`:value_#{current_index}`) just like `tag_filter_query`.
- `filter_payload` does not expand access scope — it only changes which key is read from already-permitted params.

## Known limitation (documented for follow-up)
SQL-standard operator precedence applies: in a multi-row filter that mixes AND and OR, `A OR B AND C` parses as `A OR (B AND C)`. This matches upstream Chatwoot and common filter-UI conventions. Explicit grouping (parentheses per clause) would not change precedence; honoring user-intended grouping would require a UI redesign (explicit group/nest controls).

## Test plan
- `bundle exec rspec spec/services/conversations/filter_service_spec.rb` — 4 examples, 0 failures (covers `filter_payload` key alias, nil-safe validate, and existing ordering test).
- Ruby syntax check on all changed `.rb` files — OK.
- YAML parse check on `filter_keys.yml` — OK.
- Manual: POST `/conversations/filter` with 2-row body mixing `status=snoozed` OR `inbox_id=<uuid>` now returns the inbox matches; single-filter `pipeline_id=<uuid>` returns conversations on that pipeline; multi-filter AND combinations still work.

## Changed Files
- `lib/filters/filter_keys.yml` — `pipeline_id` + `contact_id` entries under `conversations:`.
- `app/services/filter_service.rb` — `filter_payload` helper, `query_builder` refactor, new `pipeline_filter_query`.
- `app/helpers/filters/filter_helper.rb` — `'pipeline'` data_type dispatch.
- `spec/services/conversations/filter_service_spec.rb` — alias + guard specs.

## Related PRs
- Frontend: evo-ai-frontend-community PR to be linked after creation (same `fix/EVO-974` branch).

## Linked Issue
- [EVO-974](https://linear.app/evoai/issue/EVO-974)

### DIFF
--- a/app/helpers/filters/filter_helper.rb
+++ b/app/helpers/filters/filter_helper.rb
@@ -61,6 +61,8 @@ module Filters::FilterHelper
       date_filter(current_filter, query_hash, filter_operator_value)
     when 'labels'
       tag_filter_query(query_hash, current_index)
+    when 'pipeline'
+      pipeline_filter_query(query_hash, current_index)
     when 'text_case_insensitive'
       text_case_insensitive_filter(query_hash, filter_operator_value)
     else
--- a/app/services/filter_service.rb
+++ b/app/services/filter_service.rb
@@ -106,6 +106,28 @@ class FilterService
     ]
   end
 
+  def pipeline_filter_query(query_hash, current_index)
+    table_name = filter_config[:table_name]
+    query_operator = query_hash[:query_operator]
+    @filter_values["value_#{current_index}"] = query_hash['values']
+
+    base_relation_query =
+      "SELECT 1 FROM pipeline_items WHERE pipeline_items.conversation_id = #{table_name}.id"
+    id_filter =
+      "AND pipeline_items.pipeline_id IN (:value_#{current_index})"
+
+    case query_hash[:filter_operator]
+    when 'equal_to'
+      "EXISTS (#{base_relation_query} #{id_filter}) #{query_operator}"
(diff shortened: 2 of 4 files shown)


nickoliveira23:
Linked frontend PR: https://github.com/EvolutionAPI/evo-ai-frontend-community/pull/18

PR #5, opened 2026-04-19
fix(evolution_go): persist api_url and admin_token from GlobalConfig into provider_config

andersonlemesc:
## Summary

- When `EVOLUTION_GO_API_URL` and `EVOLUTION_GO_ADMIN_SECRET` are set in Admin Settings, the frontend intentionally omits these values from API requests (security design). The backend resolved them via `GlobalConfigService` to create the Evolution Go instance but never persisted them into `channel_whatsapp.provider_config`, causing all subsequent calls (`connect`, `qrcode`, `fetch`) to fail with `400 Bad Request`.
- Adds `before_validation` callback `merge_evolution_go_global_config` on `Channel::Whatsapp` that fills in `api_url` and `admin_token` from GlobalConfig when blank in `provider_config`, without overwriting explicitly provided values.
- Fixes `webhook_url` helper in `EvolutionGoConcern` to fall back to `GlobalConfigService.load('BACKEND_URL')` when `ENV['BACKEND_URL']` is unset.

Fixes EvolutionAPI/evo-crm-community#29

## Test plan

- [ ] Set `EVOLUTION_GO_API_URL` and `EVOLUTION_GO_ADMIN_SECRET` in Admin Settings → Channels → Evolution Go
- [ ] Create a new Evolution Go channel through the UI
- [ ] Confirm `provider_config` contains `api_url` and `admin_token` in the database
- [ ] Click **Connect** → QR Code generates without errors
- [ ] Run model specs: `bundle exec rspec spec/models/channel/whatsapp_spec.rb`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Persist Evolution Go WhatsApp provider configuration values from global settings and ensure webhook URLs fall back to configurable backend URLs.

Bug Fixes:
- Populate Evolution Go WhatsApp channel api_url and admin_token from GlobalConfig when missing in provider_config so subsequent operations succeed.
- Resolve webhook URL generation for Evolution Go by falling back to a stored BACKEND_URL when the environment variable is not set.

Tests:
- Add model specs for Channel::Whatsapp to cover Evolution Go provider configuration merging from global settings.

PR #14, opened 2026-04-24
fix(crm): resolve label UUIDs when tagging and rendering conversations (EVO-1001)

nickoliveira23:
## Summary

Fixes two label-related defects on the conversations view that shared the same root cause (UUIDs leaking into `cached_label_list`):

- **Activity message renders UUID** — when a client sends label identifiers as UUIDs to `POST /conversations/:id/labels`, `acts_as_taggable_on` stores the UUID verbatim as `tags.name`, which then bubbles up into the activity timeline.
- **Labels disappear from conversation cards** — for legacy rows already storing UUIDs, `ConversationSerializer` could not match the tag against `labels_by_title` and returned `labels: []`, so `ConversationBadges.tsx` rendered nothing.

## Changes

- `app/controllers/concerns/label_concern.rb` — `#create` runs `resolve_label_titles` which converts UUID-shaped inputs to their `Label#title` before `update_labels`. Non-UUID strings pass through untouched for back-compat with callers that already send titles.
- `app/serializers/conversation_serializer.rb` — `serialize` accepts `labels_by_id:` in addition to `labels_by_title:` and resolves tags via `title_index[tag.downcase] || id_index[tag]`, so legacy UUID-stored conversations keep rendering `{id, title, color}` objects.
- `app/controllers/api/v1/conversations_controller.rb` — computes both indexes in a single `Label.all.to_a` pass (request-memoised) and passes both to every `ConversationSerializer` call with `include_labels: true` (`#index`, `#search`, `#filter`, `#show`, `#create`, `#update`, `#update_custom_attribute`, `#custom_attributes`). Previously only the collection endpoints passed `labels_by_title`, so single-conversation endpoints also returned `labels: []`.

## Validation

- `ruby -c app/controllers/api/v1/conversations_controller.rb` — Syntax OK
- `ruby -c app/controllers/concerns/label_concern.rb` — Syntax OK
- `ruby -c app/serializers/conversation_serializer.rb` — Syntax OK
- `bundle exec rubocop app/controllers/concerns/label_concern.rb` — 0 offenses (other files have pre-existing offenses in lines not touched by this PR)
- `RAILS_ENV=test bundle exec rspec spec/serializers/conversation_serializer_spec.rb spec/requests/api/v1/labels_spec.rb spec/requests/api/v1/conversations_pin_archive_spec.rb spec/services/conversations/filter_service_spec.rb` — 23 examples, 0 failures

## Changed Files

- `app/controllers/concerns/label_concern.rb`
- `app/serializers/conversation_serializer.rb`

## Linked Issue

- EVO-1001 — https://linear.app/evoai/issue/EVO-1001

## Related PRs

- Frontend counterpart: will be linked after creation

## Summary by Sourcery

Normalize and correctly resolve conversation labels so both legacy UUID-stored tags and new title-based tags render consistently across all conversation endpoints.

Bug Fixes:
- Ensure labels applied via the labels API are stored and displayed using human-readable titles instead of leaking UUIDs into activity messages and conversation views.
- Restore label badges on conversation cards for legacy conversations whose cached labels were stored as UUIDs rather than titles across all conversation endpoints.

Enhancements:
- Add centralized label title resolution in the shared label concern to transparently convert UUID inputs to label titles while preserving backward compatibility for title-based callers.
- Extend the conversation serializer and controller to use both title- and ID-based label indexes so label metadata is correctly attached whenever labels are included in responses.

### DIFF
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -40,7 +40,8 @@ class Api::V1::ConversationsController < Api::V1::BaseController
         include_labels: true,
         unread_counts: unread_counts_map(conversation_ids),
         last_non_activity_messages: last_non_activity_messages_map(conversation_ids),
-        labels_by_title: labels_by_title
+        labels_by_title: labels_by_title,
+        labels_by_id: labels_by_id
       ),
       meta: conversations_pagination_meta,
       message: 'Conversations retrieved successfully'
@@ -70,7 +71,8 @@ class Api::V1::ConversationsController < Api::V1::BaseController
         include_labels: true,
         unread_counts: unread_counts_map(conversation_ids),
         last_non_activity_messages: last_non_activity_messages_map(conversation_ids),
-        labels_by_title: labels_by_title
+        labels_by_title: labels_by_title,
+        labels_by_id: labels_by_id
       ),
       meta: conversations_pagination_meta,
       message: 'Conversations search completed successfully'
@@ -110,7 +112,13 @@ class Api::V1::ConversationsController < Api::V1::BaseController
 
   def show
     success_response(
-      data: ConversationSerializer.serialize(@conversation, include_messages: true, include_labels: true),
+      data: ConversationSerializer.serialize(
+        @conversation,
+        include_messages: true,
+        include_labels: true,
+        labels_by_title: labels_by_title,
+        labels_by_id: labels_by_id
+      ),
       message: 'Conversation retrieved successfully'
     )
   end
@@ -121,7 +129,13 @@ class Api::V1::ConversationsController < Api::V1::BaseController
       Messages::MessageBuilder.new(Current.user, @conversation, params[:message]).perform if params[:message].present?
       
       success_response(
-        data: ConversationSerializer.serialize(@conversation, include_messages: true, include_labels: true),
+        data: ConversationSerializer.serialize(
+          @conversation,
+          include_messages: true,
+          include_labels: true,
+          labels_by_title: labels_by_title,
+          labels_by_id: labels_by_id
+        ),
         message: 'Conversation created successfully',
         status: :created
       )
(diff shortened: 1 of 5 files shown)


DavidsonGomes (approved):
LGTM after the review-driven follow-up commit on top of the original PR.

**HIGH-1 fixed (AC #5)** — added regression coverage for the new behaviour:
- `spec/controllers/concerns/label_concern_spec.rb` — exercises `resolve_label_titles` via an anonymous host class. Covers blank/nil input, all-titles input (with assertion that no DB call happens), all-UUIDs input, mixed input with dedupe, and unresolvable UUIDs (silently dropped, non-UUID entries preserved).
- `spec/serializers/conversation_serializer_spec.rb` — adds an `include_labels: true` context. Pins: tag stored as title resolves via `labels_by_title`; tag stored as legacy UUID resolves via `labels_by_id` fallback; `title_index` wins over `id_index` when both contain the tag; unresolvable tags drop; `include_labels: false` keeps `labels` as `[]`.

This is the regression test AC #5 explicitly required. Without it the next ill-considered touch on `LabelConcern` or `ConversationSerializer` reintroduces the UUID-leak bug silently.

**Follow-ups (out of scope, file Linear tickets):**
- **MED-1: Macros + legacy automation rules still UUID-leak.** `ActionService#add_label` (used by `Macros::ExecutionService` and `AutomationRules::ActionService` when `rule.mode != 'flow'`) calls `add_labels(labels)` without UUID resolution. The newer `AutomationRules::FlowExecutionService` (`flow_execution_service.rb:220`) already resolves correctly via `Label.where(id: ids).pluck(:title)` — the legacy path just hasn't adopted that pattern. Same root cause; scope-adjacent.
- **LOW-1: Legacy data first-edit produces noisy "removed" activity.** For conversations with UUIDs already in `cached_label_list`, the first label edit's diff lists those legacy UUIDs as the "removed" entries since the diff is computed against the pre-write list. Acceptable as a one-off migration artifact, but AC #3 reads strictly. Consider a one-time backfill job that rewrites `cached_label_list` for affected rows.
- **LOW-2: Sourcery's "comma-separated string" concern.** Mitigated today by `permit(labels: [])` in both controllers including `LabelConcern`, but a defensive `if labels.is_a?(String) ... split(',')` on the resolver would prevent the silent bypass if anyone widens the param shape later.

Full review notes: `workspace/development/features/evo-1001/[C]code-review-evo-1001.md`.

PR #22, opened 2026-04-26
fix(evolution-go): fix conversation routing, media display and ActionCable stability

andersonlemesc:
## Context

EvoGo (`evolution-go`) is the Go-based WhatsApp provider. Several integration issues prevented basic functionality: outgoing messages created duplicate conversations, media messages were saved with empty content (no attachment), and ActionCable broadcasts could fail silently. All fixes in this PR are required together for a working EvoGo integration.

### 1. Conversation routing — no more duplicate conversations
**Root cause:** When the CRM sends a message via EvoGo, EvoGo echoes it back as an incoming webhook with `IsFromMe: true`. The service was looking up the contact by phone number but outgoing messages use a LID (`@lid`) identifier, so no match was found and a new conversation was created on every send.

**Fix:** `find_or_create_conversation` now looks up the contact inbox by LID identifier first, then falls back to phone number. Outgoing echo messages are routed to the existing conversation instead of creating a new one.

**Files:** `incoming_message_evolution_go_service.rb`, `messages_upsert.rb`

---

### 2. Contact lookup and outgoing message type
**Root cause:** Contact lookup query was using the wrong join condition and outgoing messages were being assigned `sender_type: Contact` instead of `sender_type: User`, causing incorrect conversation attribution.

**Fix:** Correct inbox join in contact lookup; set outgoing sender to the inbox owner (User). Pending conversations are also reopened when a new incoming message arrives.

**Files:** `messages_upsert.rb`

---

### 3. Media messages not displaying
Three separate root causes, all required:

**a) ActiveStorage `after_commit` not firing in Sidekiq**
Webhook processing runs inside `Webhooks::WhatsappEventsJob` (Sidekiq). ActiveStorage's `after_commit` callback does not fire in this context, so blobs were enqueued but never written — the attachment record existed but had no file.

**Fix:** Replace `attachment.file.attach(blob)` deferred pattern with `ActiveStorage::Blob.create_and_upload!` which writes synchronously.

**b) `mediaUrl` nested inside message sub-type**
EvoGo may place `mediaUrl` at the root `Message` level or nested inside `imageMessage`, `audioMessage`, `videoMessage`, etc. `download_attachment_file` only checked the root level, so nested URLs were missed.

**Fix:** `extract_media_url` checks root first, then falls back to each sub-type key.

**c) EvoGo without S3 sends `base64` instead of `mediaUrl`**
When EvoGo is not configured with S3/object storage, it encodes the media as base64 and sends it inline in `Message.base64`. No URL is present.

**Fix:** If `extract_media_url` returns nil, decode `Message.base64` into a `Tempfile` and use that as the attachment source. S3-configured instances are unaffected (URL path is tried first).

**Files:** `messages_upsert.rb`

---

### 4. Duplicate method definitions silencing audio metadata
**Root cause:** `configure_audio_metadata` and `audio_voice_note?` were each defined twice in the same module. Ruby silently uses the **last** definition, discarding the first. The last definitions were incomplete stubs that used wrong key types (string keys on a symbol-keyed hash), so waveform, duration, and PTT metadata were never set.

**Fix:** Merge into single correct definitions using symbol keys, covering all metadata fields (`ptt`, `seconds`, `fileLength`, `waveform`).

Also removed two dead methods (`save_message_and_notify`, `attach_media_from_url`) that were never called anywhere.

**Files:** `messages_upsert.rb`

---

### 5. ActionCable stability
**Root cause 1:** `ActionCableBroadcastJob` received `data` with string keys (`data['id']`) but some event paths produced symbol keys (`data[:id]`), causing `conversation_id` lookup to return nil and `Conversation.find_by!` to raise.

**Root cause 2:** `ActionCableListener` built token arrays with `[account_token(account)]` where `account_token` could return `""` (empty string) when account was nil, broadcasting to an empty-string channel.

**Fix:** Handle both key types in `ActionCableBroadcastJob`; `account_token` returns `nil` (not `""`) and all token arrays use `.compact`.

**Files:** `action_cable_broadcast_job.rb`, `action_cable_listener.rb`

---

## Test plan

- [ ] Send text message via WhatsApp → appears once in conversation (no duplicate)
- [ ] Reply from CRM → EvoGo echo routes to same conversation, not a new one
- [ ] Send image via WhatsApp (EvoGo without S3) → image displays in conversation
- [ ] Send audio/PTT via WhatsApp → audio player shows with correct duration and waveform
- [ ] Send video/document via WhatsApp → attachment displays correctly
- [ ] Pending conversation receives incoming message → status changes to open
- [ ] ActionCable broadcasts do not raise errors in Sidekiq logs

### DIFF
--- a/app/jobs/action_cable_broadcast_job.rb
+++ b/app/jobs/action_cable_broadcast_job.rb
@@ -27,7 +27,8 @@ class ActionCableBroadcastJob < ApplicationJob
   def prepare_broadcast_data(event_name, data)
     return data unless CONVERSATION_UPDATE_EVENTS.include?(event_name)
 
-    conversation = Conversation.find_by!(id: data[:id])
+    conversation_id = data[:id] || data['id']
+    conversation = Conversation.find_by!(id: conversation_id)
     conversation.push_event_data
   end
 
--- a/app/listeners/action_cable_listener.rb
+++ b/app/listeners/action_cable_listener.rb
@@ -33,7 +33,7 @@ class ActionCableListener < BaseListener
   def message_created(event)
     message, account = extract_message_and_account(event)
     conversation = message.conversation
-    tokens = user_tokens(account, conversation.inbox.members) + contact_tokens(conversation.contact_inbox, message) + [account_token(account)]
+    tokens = (user_tokens(account, conversation.inbox.members) + contact_tokens(conversation.contact_inbox, message) + [account_token(account)]).compact
 
     broadcast(account, tokens, MESSAGE_CREATED, message.push_event_data)
   end
@@ -41,7 +41,7 @@ class ActionCableListener < BaseListener
   def message_updated(event)
     message, account = extract_message_and_account(event)
     conversation = message.conversation
-    tokens = user_tokens(account, conversation.inbox.members) + contact_tokens(conversation.contact_inbox, message) + [account_token(account)]
+    tokens = (user_tokens(account, conversation.inbox.members) + contact_tokens(conversation.contact_inbox, message) + [account_token(account)]).compact
 
     broadcast(account, tokens, MESSAGE_UPDATED, message.push_event_data.merge(previous_changes: event.data[:previous_changes]))
   end
@@ -56,7 +56,7 @@ class ActionCableListener < BaseListener
 
   def conversation_created(event)
     conversation, account = extract_conversation_and_account(event)
-    tokens = user_tokens(account, conversation.inbox.members) + contact_inbox_tokens(conversation.contact_inbox) + [account_token(account)]
+    tokens = (user_tokens(account, conversation.inbox.members) + contact_inbox_tokens(conversation.contact_inbox) + [account_token(account)]).compact
 
     broadcast(account, tokens, CONVERSATION_CREATED, conversation.push_event_data)
   end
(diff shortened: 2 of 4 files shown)


DavidsonGomes:
Anderson, este PR ficou conflitante com `develop` depois que mergeamos a Onda 2 hoje (#34, #35, #37 — esta última toca `action_cable_listener.rb` e os services de Contact via SyncDispatcher).

Pode fazer rebase? O conteúdo do seu PR está aprovado conceitualmente (corrige 5 bugs reais do EvoGo: routing duplicado, mídia sem arquivo, áudio sem metadata, ActionCable instável, sender_type errado) — só precisamos resolver os conflitos com a develop atual.

Pontos prováveis de conflito:
- `app/listeners/action_cable_listener.rb` (#37 mexeu aqui também, mesmo padrão de `.compact` + nil-handling)
- Outros listeners se você ajustou padrões similares

Quando rebasear, comenta aqui que eu reviso e mergeio.

DavidsonGomes:
Merged via cherry-pick. Os 9 commits foram aplicados na develop com a autoria do @andersonlemesc preservada (commits b700ae6 → cde3adc). Resolvi um conflito em app/listeners/action_cable_listener.rb mantendo a versão mais robusta do account_token (que aceita tanto Hash quanto AR-object). Obrigado pelo trabalho!

PR #68, opened 2026-05-15
fix(crm): apply storage provider from admin UI at runtime [EVO-1050]

marcelogorutuba:
Applies the storage provider chosen in Admin Settings → Storage at runtime in the CRM service, without requiring a process restart.

## What changed

- **`config/initializers/active_storage_dynamic_service.rb`**: Overrides `ActiveStorage::Blob.service` via `class_eval` + `alias_method` so the service is resolved dynamically from `GlobalConfigService` on every call instead of being fixed at boot. Guard added (`next if respond_to?(:_static_service)`) to prevent infinite recursion on initializer double-load (dev reload, Zeitwerk, class-cache restart without process reboot).
- **`config/environments/production.rb`**: Boot-time fallback refactored from inline rescue to explicit `begin/rescue` block with `warn` log so failures at startup are traceable.

## Operational notes

- **S3 credentials** (access key, secret, region, endpoint) are still read at boot from `config/storage.yml` ERB — changing them via the UI requires restarting the service and its Sidekiq workers.
- **Provider changes** (Local ↔ S3 ↔ S3-Compatible) take effect immediately after save (cache invalidated via `GlobalConfig.set`). Auth converges within **60 seconds** (passive TTL); there is a window of up to 60 s where auth and CRM may use different providers — the storage UI banner communicates this to operators.
- **Existing blobs** remain on the original provider; only new uploads use the updated configuration. Migration/re-attachment is explicitly out of scope.
- ENV vars remain a valid override/fallback for infra-as-code installs (AC #3).

## Validation

- `ruby -c config/initializers/active_storage_dynamic_service.rb` ✅
- `ruby -c config/environments/production.rb` ✅

## Linked Issue

- EVO-1050

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -33,7 +33,12 @@ Rails.application.configure do
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options)
-  config.active_storage.service = (GlobalConfigService.load('ACTIVE_STORAGE_SERVICE', ENV.fetch('ACTIVE_STORAGE_SERVICE', 'local')) rescue ENV.fetch('ACTIVE_STORAGE_SERVICE', 'local')).to_sym
+  config.active_storage.service = begin
+    GlobalConfigService.load('ACTIVE_STORAGE_SERVICE', ENV.fetch('ACTIVE_STORAGE_SERVICE', 'local'))
+  rescue StandardError => e
+    warn "[ActiveStorage] GlobalConfigService unavailable at boot (#{e.message}); falling back to ENV"
+    ENV.fetch('ACTIVE_STORAGE_SERVICE', 'local')
+  end.to_sym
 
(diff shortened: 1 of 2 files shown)


dpaes (approved):
## ✅ Approved — third-pass code review

All blockers from previous reviews are resolved:

- **Infinite-recursion guard on initializer double-load** — `next if ActiveStorage::Blob.respond_to?(:_static_service)` correctly short-circuits the override on reload.
- **Cross-service 60s window** — documented in the storage UI banner.
- **Inline rescue in `config/environments/production.rb`** — refactored to explicit `begin/rescue` with `warn` log.
- **`Rails.logger.warn` on fallback** — both the unresolved-service path and the rescue path emit a log line.

### Non-blocking follow-up (not gating this merge)

- **M6 — Misleading log message in fallback path** — `config/initializers/active_storage_dynamic_service.rb:25-28`: the `unless resolved` log claims `service '#{service_name}' not registered` even when the real reason was `respond_to?(:services)` returning false. Either differentiate the two messages or drop the `respond_to?(:services)` check (always true post-`after_initialize`).

Merging and moving the card to Done.

PR #87, opened 2026-05-19
feat(crm): EvoFlow module foundation — client + payload + worker (EVO-1238)

daniloleonecarneiro:
### What's in
- **`EvoFlow::Client`** (`app/services/evo_flow/client.rb`) — instance/DI HTTParty client, `X-Integration-API-Key`, configurable timeout, relative-URI join (keeps `/api/v1`), parse-safe, redacted auth logs. `+ EvoFlow::HTTPError / EvoFlow::ConfigurationError`.
- **`EvoFlow::PayloadBuilder`** — real evo-flow DTOs (camelCase, single-tenant: no `accountId`/`eventType`), deterministic `message_id_for` (SHA256), UTC ISO-8601 with validation.
- **`EvoFlow::PublishEventWorker`** — pure Sidekiq (`queue: :integrations, retry: 5`), `sidekiq_retries_exhausted` → Wisper `:evo_flow_publish_failed`, PII (`properties`/`traits`) redacted before Redis/Dead Set/broadcast.
- **`lib/events/evo_flow_event_names.rb`** — frozen `EvoFlow::EVENT_NAMES` (16 events) + Zeitwerk ignore/require wiring in `config/application.rb` (constant name ≠ path).
- **`.env.example`** — `EVO_FLOW_API_URL`, `AUTH_APIKEY_INTEGRATION_LOCAL`, `EVO_FLOW_ALLOW_INSECURE`.

### Documented divergences
- `Wisper.publish` does not exist in Wisper 2.0.0 → internal `FailureBroadcaster` (`include Wisper::Publisher`), same event/payload, repo idiom.
- `AUTH_APIKEY_INTEGRATION_LOCAL` is **created** here; worker is `EvoFlow::PublishEventWorker` in `app/workers/`; `/api/v1` lives in `EVO_FLOW_API_URL`; payload uses real DTO; `perform(path, payload)`.

### Quality
- Adversarial review: **13/13 findings resolved** (TLS guard, parse-safety, log/PII redaction, uniform retry path, ENV consistency, URL-join hardening, ISO-8601 validation, blank-key fail-fast, integration + eager-load regression specs).
- RuboCop clean (EvoFlow files). **RSpec 33/33**. `zeitwerk:check` OK. `Rails.application.eager_load!` OK.

### Known caveat (non-blocking, pre-existing in spec)
evo-flow has no consumer-side dedup yet (`contact_events` MergeTree); Sidekiq retries still duplicate downstream until a future evo-flow story. `message_id_for` is the correct forward-looking foundation.

## Summary by Sourcery

Introduce the foundational EvoFlow CRM integration with a dedicated HTTP client, payload builder, Sidekiq worker, and canonical event name list, wired into Rails autoloading.

New Features:
- Add EvoFlow::Client service for authenticated, parse-safe HTTP communication with the evo-flow API, including configuration validation and secure error handling.
- Add EvoFlow::PayloadBuilder to construct evo-flow track and identify payloads with deterministic message IDs and normalized timestamps.
- Add EvoFlow::PublishEventWorker Sidekiq worker to enqueue EvoFlow publishing jobs with retries, logging, and Wisper-based failure broadcasting.
- Define EvoFlow::EVENT_NAMES as the canonical list of CRM events emitted to evo-flow.

Enhancements:
- Configure Rails autoloader to ignore and explicitly require the EvoFlow event names file so the constant is always available and eager loading remains stable.

Tests:
- Add unit and integration specs for the EvoFlow client, payload builder, event names, and publish event worker to validate behaviour, retries, and error handling.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -257,6 +257,15 @@ USE_OPTIMIZED_CONVERSATION_FINDER=false
 # Internal API token for secure communication between EvoAI services
 EVOAI_CRM_API_TOKEN=
 
+# EvoFlow Integration (event tracking → evo-flow)
+# Base URL INCLUDING the /api/v1 prefix
+EVO_FLOW_API_URL=http://evo-flow:3000/api/v1
+# Shared service-to-service key; MUST equal evo-flow's AUTH_APIKEY_INTEGRATION_LOCAL (no default)
+AUTH_APIKEY_INTEGRATION_LOCAL=
+# In production the client refuses to send the key over plain http. Set to
+# 'true' ONLY when evo-flow is reached over a trusted private network.
+EVO_FLOW_ALLOW_INSECURE=
+
 # Fernet encryption key for sensitive configuration values (e.g., API keys stored in installation_configs)
 # Must be a Fernet-compatible base64-encoded 256-bit key
 # Generate with: ruby -e "require 'securerandom'; require 'base64'; puts Base64.urlsafe_encode64(SecureRandom.random_bytes(32))"
--- a/app/services/evo_flow/client.rb
+++ b/app/services/evo_flow/client.rb
@@ -1,0 +1,102 @@
+module EvoFlow
+  # Raised on any non-2xx evo-flow response, an unparseable body, or a network
+  # failure. Mirrors Crm::Hubspot::Api::BaseClient::ApiError (code + response).
+  class HTTPError < StandardError
+    attr_reader :code, :response
+
+    def initialize(message = nil, code = nil, response = nil)
+      @code = code
+      @response = response
+      super(message)
+    end
+  end
+
+  # Raised at construction time for an unusable configuration (missing key, or
+  # cleartext transport in production). Fails fast instead of emitting a
+  # request that is guaranteed to 401 or that leaks the shared key in cleartext.
+  class ConfigurationError < StandardError; end
+
+  # Instance-based (DI-friendly) authenticated HTTP client for evo-flow.
+  # Pattern mirrors app/services/crm/hubspot/api/base_client.rb (HTTParty +
+  # custom error + handle_response).
+  class Client
+    include HTTParty
+
+    DEFAULT_API_URL = 'http://evo-flow:3000/api/v1'.freeze
+    REDACTED_BODY = '[redacted: auth failure]'.freeze
+    MAX_LOGGED_BODY = 500
+
+    def initialize(api_url: ENV.fetch('EVO_FLOW_API_URL', DEFAULT_API_URL),
+                   api_key: ENV.fetch('AUTH_APIKEY_INTEGRATION_LOCAL', nil),
+                   timeout: 10)
+      @api_url = api_url
+      @api_key = api_key
+      @timeout = timeout
+      validate_config!
+    end
+
+    def post(path, payload)
+      response = self.class.post(join(@api_url, path),
+                                 body: payload.to_json,
+                                 headers: request_headers,
+                                 timeout: @timeout)
+      handle_response(response)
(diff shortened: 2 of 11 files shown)


dpaes (approved):
LGTM — approving.

Adversarial review covered AC validation, security/PII handling, error paths, test quality and Zeitwerk wiring. Six ACs are met (AC5/AC1 default URL/worker naming/perform signature diverge from the literal story text but match the *real* evo-flow contract — verified against `evo-flow/src/modules/events/dto/track-event.dto.ts`). The AC text in Linear should be updated to match the real DTO so future devs don't try to "fix" the code back toward stale spec.

### Strengths
- DI-friendly `EvoFlow::Client`, mirrors the existing `Crm::Hubspot::Api::BaseClient` pattern.
- TLS guard + cleartext refusal in production with an explicit `EVO_FLOW_ALLOW_INSECURE` opt-out.
- Auth-rejection bodies (401/403) redacted before logs/error messages — closes a real key-leak vector.
- Deterministic `message_id_for(event, contact, source_uuid)` — correct forward-looking idempotency foundation.
- `FailureBroadcaster` is the right workaround for Wisper 2.0.0's missing `Wisper.publish`.
- Zeitwerk ignore + explicit require for `EvoFlow::EVENT_NAMES` is correct, with a regression spec calling `Rails.application.eager_load!`.
- Integration spec exercises `PayloadBuilder → Worker → Client → wire` with only the HTTP boundary mocked.

### Non-blocking findings (informational, recorded on EVO-1238 in Portuguese)
- **MEDIUM** — Asymmetric PII redaction in `sidekiq_retries_exhausted`: payload is sanitized but `ex.message` is forwarded raw to logs and the Wisper broadcast. For non-HTTPError terminals (parse errors, future TLS errors) the raw message may leak response excerpts. Consider a `sanitize_error` helper.
- **MEDIUM** — `validate_config!` only rejects non-`https` schemes in production. A typo like `EVO_FLOW_API_URL=evo-flow:3000` (no scheme) silently validates in dev and fails unhelpfully at first request. Suggest rejecting `scheme not in [http, https]` at boot regardless of env.
- **MEDIUM** — AC4's Wisper `:evo_flow_publish_failed` has zero subscribers in production (grep returns only the worker + spec). Structurally satisfied, operationally dead until a follow-up story wires a listener. Terminal-failure monitoring relies on the Sidekiq Dead Set until then — worth calling out in the merge note.
- **MEDIUM** — Out-of-scope edit in `.env.example`: the SMTP-section comment rewrite is unrelated to EVO-1238 and should ride a chore PR.
- **LOW** — `EVO_FLOW_ALLOW_INSECURE == 'true'` is strict-string; doc should make that explicit (`'1'`/`'TRUE'`/`'yes'` fail silently).
- **LOW** — `safe_body` only redacts 401/403; 4xx with body echo (400/422/429) may still leak reflected payloads. Consider redacting all 4xx by default.
- **LOW** — `iso8601(nil)` falls back to `Time.current` silently — masks a missing `occurred_at`. Raise instead and force the caller to be explicit.
- **LOW** — Rescue list in `Client#post` may not cover `Net::OpenTimeout` and `OpenSSL::SSL::SSLError`. WebMock's `to_timeout` raises `Timeout::Error` so the spec passes, but a real misconfigured-host smoke test would confirm those exceptions are wrapped, not raw-raised to Sidekiq.
- **LOW** — `# frozen_string_literal: true` missing on the four new EvoFlow files (repo is inconsistent — nit).
- **LOW** — `message_id(payload)` logs an empty string when both `'messageId'`/`:messageId` are absent; consider `'<missing>'` so log grepping works.

These are recorded on Linear without follow-up tickets per Daniel's call.

PR #94, opened 2026-05-25
feat(crm): publish Wisper :message_status_changed for all channels (EVO-1239)

daniloleonecarneiro:
## Summary

- Funnels every `Message#status` write through `Messages::StatusUpdateService` so the canonical Wisper payload `{message, previous_status, status, external_error}` fires for EvoFlow tracking. **8 previously-bypassing paths now delegate, 2 new emissions added, 1 new mailbox handles inbound DSN bounces.**
- WhatsApp (all providers): Cloud + 360-dialog, Evolution API, Evolution Go (`messages.update` + single/bulk receipts), Baileys — all funnel through the canonical service. Bulk receipts keep `update_all` for performance and emit per-row via a dedicated `BulkStatusPublisher`. Baileys-specific rules preserved (`read` is final, `delivered → sent` rejected).
- Telegram: `delivered` is now emitted after a successful send; `failed` keeps the `"<code>, <description>"` external_error format. `read` is permanently n/a (Bot API limitation for general chats).
- Email/SMTP: `EmailReplyWorker` captures the outbound `Message-Id` into `source_id` and emits `delivered` after `deliver_now`; rescue is scoped to `deliver_now` via a private helper so post-send failures cannot mark a delivered email failed. New `BounceMailbox` parses RFC 3464 multipart/report DSNs (`5.x.x` → `failed`, `4.x.x` logged only). `ApplicationMailbox` routes DSN-content emails to `BounceMailbox` **before** the reply route would catch them.

**Scope expansion vs. original ticket:** the original "Escopo — fora" assumed Evolution API WhatsApp already emitted `:message_status_changed`. Step-2 investigation falsified this — Evolution API, Evolution Go, and Baileys all wrote `Message#status` directly. Confirmed with user on 2026-05-25 to instrument all WhatsApp providers. See the bypass-site table in the Linear comment for the full mapping.

**Blocks:** EVO-1240 (story 7.4 — EvoFlow listener port). EVO-1239 must merge first.

## Test plan

- [ ] CI green on `develop` base.
- [ ] **Local rspec:** 47 examples / 0 failures across 9 new + 2 extended specs:
  - [ ] `spec/services/whatsapp/incoming_message_base_service_spec.rb` — delivered/read/failed + external_error format `"<code>: <title>"`
  - [ ] `spec/services/whatsapp/evolution_handlers/messages_update_spec.rb`
  - [ ] `spec/services/whatsapp/evolution_go_handlers/messages_update_spec.rb` — includes same-status guard
  - [ ] `spec/services/whatsapp/evolution_go_handlers/receipt_handler_spec.rb` — single + `BulkStatusPublisher`
  - [ ] `spec/services/whatsapp/baileys_handlers/messages_update_spec.rb` — `delivered → sent` rejection + read-finality
  - [ ] `spec/services/telegram/send_on_telegram_service_spec.rb` — success / nil-id / already-failed
  - [ ] `spec/models/channel/telegram_spec.rb` — `#process_error` comma-format
  - [ ] `spec/workers/email_reply_worker_spec.rb` — delivered + msg-id capture + rescue
  - [ ] `spec/mailboxes/bounce_mailbox_spec.rb` — DSN parse (permanent / transient / no-match) + routing-order via `ApplicationMailbox.router.mailbox_for`
- [ ] **Local rubocop:** 0 novas offenses (18 pré-existentes nos arquivos tocados, padrões do codebase).
- [ ] **Smoke manual:** outbound WhatsApp Cloud → ver Wisper `delivered`/`read` no log; outbound Telegram → ver `delivered`; outbound Email → ver `delivered`; trigger bounce contra `invalid@nonexistent-12345.example.com` → ver `BounceMailbox` processar DSN.
- [ ] **Regression watch:** `notificame`, `zapi`, Facebook/Instagram/Line/Twilio continuam emitindo (não alterados). `Messages::StatusUpdateService` em si não foi tocado.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- N/A
+- **EVO-1239** — Wisper `:message_status_changed` agora é emitido por todos os providers WhatsApp (Cloud, 360-dialog, Evolution API, Evolution Go single/bulk, Baileys), Telegram (delivered + failed) e Email/SMTP (delivered + bounce DSN). Inclui novo `BounceMailbox` parseando DSN RFC 3464 (Status `5.x.x` → failed; `4.x.x` apenas logado).
 
 ### Changed
 
-- N/A
+- **EVO-1239** — Telegram `send_on_telegram` passa a registrar status `delivered` após envio bem-sucedido. Read receipts não são suportados pela Bot API (Telegram limitation) e portanto `read` permanece n/a neste canal.
 
 ### Fixed
 
--- a/app/controllers/api/v1/conversations/messages_controller.rb
+++ b/app/controllers/api/v1/conversations/messages_controller.rb
@@ -32,19 +32,16 @@ class Api::V1::Conversations::MessagesController < Api::V1::Conversations::BaseC
 
   def update
     @message = message
-    Messages::StatusUpdateService.new(@message, permitted_params[:status], permitted_params[:external_error]).perform
+    previous_status = @message.status
+    target_status = permitted_params[:status]
+    return invalid_transition_response(previous_status, target_status) unless perform_status_update(target_status)
 
     success_response(
       data: MessageSerializer.serialize(@message.reload, include_attachments: true, include_sender: true),
       message: 'Message updated successfully'
     )
   rescue StandardError => e
-    error_response(
-      ApiErrorCodes::VALIDATION_ERROR,
-      'Failed to update message',
-      details: e.message,
-      status: :unprocessable_entity
-    )
+    error_response(ApiErrorCodes::VALIDATION_ERROR, 'Failed to update message', details: e.message, status: :unprocessable_entity)
   end
 
   def destroy
@@ -73,10 +70,10 @@ class Api::V1::Conversations::MessagesController < Api::V1::Conversations::BaseC
   def retry
     @message = message
 
+    # Reset to :sent directly (StatusUpdateService would reject the no-op
+    # `sent → sent` Wisper publish anyway). Channel webhooks emit the
+    # subsequent delivered/read events through the canonical funnel.
     @message.update!(status: :sent, content_attributes: {})
-
-    service = Messages::StatusUpdateService.new(@message, 'sent')
-    service.perform
 
     ::SendReplyJob.perform_now(@message.id)
 
@@ -94,6 +91,19 @@ class Api::V1::Conversations::MessagesController < Api::V1::Conversations::BaseC
   end
 
   private
+
+  def perform_status_update(target_status)
+    Messages::StatusUpdateService.new(@message, target_status, permitted_params[:external_error]).perform
+  end
+
+  def invalid_transition_response(previous_status, target_status)
+    error_response(
+      ApiErrorCodes::VALIDATION_ERROR,
+      'Invalid status transition',
+      details: "#{previous_status} → #{target_status}",
+      status: :unprocessable_entity
+    )
+  end
 
   def message
     @message ||= @conversation.messages.find(permitted_params[:id])
--- a/app/mailboxes/application_mailbox.rb
+++ b/app/mailboxes/application_mailbox.rb
@@ -6,6 +6,12 @@ class ApplicationMailbox < ActionMailbox::Base
   REPLY_EMAIL_UUID_PATTERN = /^reply\+([0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12})$/i
   CONVERSATION_MESSAGE_ID_PATTERN = %r{conversation/([a-zA-Z0-9-]*?)/messages/(\d+?)@(\w+\.\w+)}
 
+  # DSN bounces are sent BY mail systems TO our outbound reply+UUID@... address.
+  # Must route to BounceMailbox BEFORE the reply route would catch them.
+  routing(
+    ->(inbound_mail) { delivery_status_notification?(inbound_mail) } => :bounce
+  )
+
   # routes as a reply to existing conversations
   routing(
     ->(inbound_mail) { valid_to_address?(inbound_mail) && (reply_uuid_mail?(inbound_mail) || in_reply_to_mail?(inbound_mail)) } => :reply
@@ -20,6 +26,13 @@ class ApplicationMailbox < ActionMailbox::Base
   routing(all: :default)
 
   class << self
+    # Matches RFC 3464 delivery-status notifications regardless of whose
+    # address they were sent TO (often our outbound reply+UUID@... address).
+    def delivery_status_notification?(inbound_mail)
+      ct = inbound_mail.mail.content_type.to_s
+      ct.include?('multipart/report') && ct.include?('report-type=delivery-status')
+    end
+
     # checks if follow this pattern then send it to reply_mailbox
     # <account/#{@account.id}/conversation/#{@conversation.uuid}@#{@account.inbound_email_domain}>
     def in_reply_to_mail?(inbound_mail)
--- a/app/mailboxes/bounce_mailbox.rb
+++ b/app/mailboxes/bounce_mailbox.rb
@@ -1,0 +1,52 @@
+class BounceMailbox < ApplicationMailbox
+  # Parses RFC 3464 multipart/report; report-type=delivery-status messages,
+  # looks up the original outbound Message by its Message-Id, and emits a
+  # `:failed` status (via Messages::StatusUpdateService) when the DSN is a
+  # permanent (5.x.x) failure. Transient (4.x.x) DSNs are logged only.
+  def process
+    return Rails.logger.warn('[BounceMailbox] No delivery-status part; dropping') if dsn_payload.blank?
+
+    original_message_id = extract_original_message_id
+    return Rails.logger.warn('[BounceMailbox] Missing Original-Message-ID; dropping') if original_message_id.blank?
+
+    message = Message.find_by(source_id: original_message_id)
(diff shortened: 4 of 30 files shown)


dpaes:
## Review verdict: **changes requested (blocked)** ⛔

@daniloleonecarneiro — two quality blockers below (**F-1** and **F-2**) before this can merge. The 5 ACs are functionally met and the architecture is right (centralizing on `Messages::StatusUpdateService` is the correct call, and I verified the Wisper wiring reaches the listener). The blockers are about **idempotency/finality of the status transitions** feeding `EvoFlow::MessageEventsListener` — both collapse to one central fix.

---

Reviewed at `633a2dd9` (head) vs `develop` (`dc644357`), isolated via three-dot. Branch is on top of the latest `develop` (HEAD == merge-base, `MERGEABLE`, no drift). First-source read of the 11 production files, the canonical `Messages::StatusUpdateService` (everything now delegates to it), the **consumer** `EvoFlow::MessageEventsListener` (the reason this story exists — it blocks EVO-1240), and the Wisper wiring. Canonical spec = the card's 5 ACs. Scope expansion to all WhatsApp providers is a justified correction of a wrong "out of scope" assumption (Evolution API/Go/Baileys also wrote `Message#status` directly), confirmed with the user on 2026-05-25 and documented — not scope creep.

### Solid work
- **Right architectural call: funnel through `Messages::StatusUpdateService`.** The ticket suggested a raw `Rails.configuration.dispatcher.dispatch`, but you reused the canonical service already used by 8+ channels. I verified the wiring: `EvoFlow::MessageEventsListener` is registered as a **global** Wisper subscriber (`evo_flow_listeners.rb:9-13`), so the instance-level `publish(:message_status_changed, data:)` **does** reach it — the instrumentation is live, not dead code.
- **Canonical payload `{message, previous_status, status, external_error}`** satisfies AC#5 as a superset; the listener reads exactly that shape.
- **`BulkStatusPublisher`** preserves `update_all` performance, captures `previous_status` *before* the update, mirrors the canonical contract, and publishes outside the transaction. Payload types are consistent (all strings).
- **`EmailReplyWorker`**: rescue is scoped to `deliver_now` only (private `deliver_email`), so a post-send failure cannot mark a delivered email failed — correct and well-commented. Captures `mail.message_id` into `source_id` for DSN matching.
- **`BounceMailbox` + routing precedence**: the DSN route is registered **before** the reply route (`application_mailbox.rb:9-11`), with a real test (`bounce_mailbox_spec.rb:64-85`) proving via `ApplicationMailbox.router.mailbox_for` that a DSN addressed to `reply+UUID@` wins over `ReplyMailbox`. DSN specs use real `.eml` fixtures.
- **`external_error` storage unchanged** — it's a `store :content_attributes` accessor (`message.rb:104-107`), so old `message.external_error =` and the service's `content_attributes.merge(external_error:)` write to the same place.
- **Finality preserved for Baileys and Evo Go receipts** (`baileys_handlers/messages_update.rb:33-37`; `receipt_handler.rb:151-167`).

### 🔴 Blocking findings

**F-1 — Major / the crux: same-status re-emit ships a FALSE `message.read`.**
The listener resolves the event name from **`previous_status` alone** (`message_events_listener.rb:77-81`) using `{ 'sent' => 'message.delivered', 'delivered' => 'message.read' }`. So a **duplicate `delivered` webhook** (WhatsApp Cloud / Evolution both resend status updates) lands as `delivered→delivered` → `previous_status='delivered'` → the listener maps it to **`message.read`** and ships a read event that never happened. `sent→sent` ships a false `message.delivered`.
You recognized this exact risk and added a same-status guard **only** to `evolution_go_handlers/messages_update.rb:65-67`. But WhatsApp Cloud (`incoming_message_base_service`), Evolution API (`evolution_handlers/messages_update`), Baileys (for `delivered→delivered`/`sent→sent`), and the canonical `StatusUpdateService#valid_status_transition?` itself (`status_update_service.rb:38-45`, only blocks `read→delivered`) have **no** same-status guard. And the rationale ("listener can't map → just warn") is incomplete: for `delivered→delivered` the listener *does* map — to the wrong event. Since this story exists to feed that listener (EVO-1240 is already on `develop`), the bug goes live the moment `evo_flow_enabled?` is turned on.

**F-2 — Major / silent finality regression: Evolution API & Evo Go `messages.update`.**
The strict transition tables (`read` final, `failed` final, `delivered→read` only) were removed from both handlers; now only the loose canonical rule applies (blocks just `read→delivered`). This is **inconsistent** with the sibling paths in the same PR (Baileys + Evo Go receipts keep finality) and isn't flagged. Walked through the listener mapping:
- `failed→delivered` / `failed→read`: a late success webhook **un-fails** a genuinely failed message (was blocked); listener → warn, but the DB status regresses.
- `failed→failed` (duplicate): ships **another `message.failed`** (duplicate/double-count).
- `delivered→sent` (downgrade, was blocked): `previous='delivered'` → **false `message.read`** (same root as F-1).
- `read→sent`: warn, but DB regresses `read`→`sent`.

> **F-1 and F-2 share one root and one fix:** harden `Messages::StatusUpdateService#valid_status_transition?` with (a) a same-status short-circuit (`return false if message.status.to_s == status.to_s`) and (b) `read`/`failed` finality. This covers **all** channels at once and makes the Evo Go local guard and the Baileys early-returns redundant (that redundancy is the consistency check). Verify against the other callers (`messages_controller`, `send_reply_job`, facebook/instagram/line/sms/twilio/whatsapp send services) — none set same-status or un-fail in the normal flow, so it should be safe; the one case to double-check is the manual status override via the API controller.

### Non-blocking
- **F-3 — Low / infra:** the only red check (`Verify EvoExtensionPoints contract`) fails at the **"Run RSpec standalone"** step (`spec/lib/evo_extension_points/`) with `SystemExit` during load + Postgres `relation "installation_configs" does not exist`. It is red **identically on every PR since ~2026-05-21** (EVO-1419/1262/1261/1088/1372/1431). This PR touches none of `lib/evo_extension_points/`, `EXTENSION_POINTS.md`, or schema. **Pre-existing `develop`-wide breakage, not attributable to this PR** — but you won't get a green CI signal until `develop` is fixed.
- **F-4 — Low:** DSN `Original-Message-ID` matching is fragile — many MTAs rewrite the Message-ID, so a bounce may not match and `failed` won't be recorded (`bounce_mailbox.rb:28-44`). Degrades gracefully (logs + drops). Worth documenting as a known SMTP-channel limitation.
- **F-5 — Low / edge:** `bulk_update_and_publish` runs the publish loop outside the transaction (correct), but Wisper `publish` is synchronous, so a listener raising mid-`find_each` leaves the remaining rows unpublished. Contained in practice (the 7.4 listener wraps in `rescue`).
- **L-6 — Low / tests:** the new handler specs `instance_double(Messages::StatusUpdateService, perform: true)` and assert *delegation* — they never exercise the real transition rules, which is exactly where F-1/F-2 live. So "47 examples, 0 failures" is green *because* the suite can't see the regression. Suggest at least one end-to-end spec per critical channel that drives the real `StatusUpdateService` and asserts the resulting event (or its absence on same-status).

### AC status
AC#1 ✅ (caveat F-1 on duplicates) · AC#2 ✅ · AC#3 ✅ (Telegram `delivered` via send ACK; `read` n/a documented) · AC#4 ✅ (caveat F-4) · AC#5 ✅.

**Net:** the redesign and the central-service approach are good; please resolve **F-1** and **F-2** (ideally the single `valid_status_transition?` fix) and we're in good shape. 🙏

dpaes (approved):
## Review — APPROVED ✅

Re-reviewed PR #94 at `084a8395` vs `develop` (`dc644357`, three-dot). The two blockers from the prior review (F-1 / F-2) are **fixed** by `7a637bf7` + `084a8395`. Branch clean, no conflict, merging into `develop`.

### Blockers resolved
- **F-1 (same-status re-emission → false `message.read`):** `Messages::StatusUpdateService#valid_status_transition?` now short-circuits same-status before `update!`/publish; the bulk pre-filter (`can_update_message_status?`) mirrors the funnel rules. Duplicate `delivered` webhooks no longer emit an event.
- **F-2 (lost read/failed finality):** centralized in the funnel (read final, failed final, `delivered → sent` barred); the divergent per-handler transition tables were removed. Confirmed `EvoFlow::MessageEventsListener` resolves `failed` by `status` *before* the `previous_status` map, so `delivered → failed` correctly maps to `message.failed` — which the email-bounce AC relies on (optimistic `delivered`, then DSN flips to `failed`).
- **L-6:** per-channel e2e specs now exercise the real funnel (no `StatusUpdateService` mock).

### 🟡 New finding (NON-blocking) — WhatsApp Cloud send-error bypass
`Whatsapp::Providers::BaseService#process_response` calls `handle_error` (`@message.update!(status: :failed)` directly, **no Wisper**) and returns `nil`. Since `send_on_whatsapp_service#send_session_message` guards with `if message_id == false` and `nil == false` is `false` in Ruby, the new funnel-delegation branch is **unreachable for the Cloud provider**. Same for `whatsapp_cloud_service#mark_audio_upload_failed`. Net effect: a Cloud **send** error marks `failed` but does not publish `:message_status_changed`, so once 7.4 (EVO-1240) lands the listener won't see `message.failed` on that path. For providers returning literal `false` (evolution_go, etc.) the fix works — only Cloud escapes via `nil`.

Not blocking: pre-existing, outside the 5 ACs (inbound `statuses[]` + DSN), and unrelated to F-1/F-2. Flagged to the author to assess a Cloud-API follow-up.

Minor (pre-existing, out of scope): `send_reply_job.rb:27` rescue and `channel/sms.rb:81` also write status without Wisper; the `EmailReplyWorker` happy path relies on `deliver_now` populating `message_id` (real ActionMailer sets it; the spec uses a double).

CI: only `Verify EvoExtensionPoints contract` fails — pre-existing F-3, not attributable to this PR (no `lib/evo_extension_points/` or schema changes here).

Issue #19, opened 2026-04-25
[bug] WhatsApp LID addressing: outgoing messages with remoteJid@lid are silently dropped


dvlexp:
## Summary

When an external IA/bot sends a reply via Evolution API (e.g., our Aurora assistant), the webhook delivers `messages.upsert` with `remoteJid: "<id>@lid"` and `addressingMode: "lid"` instead of `@s.whatsapp.net`. WhatsApp's new "LID addressing mode" (2024+ privacy feature) is used for bot/multi-device sessions.

`Whatsapp::EvolutionHandlers::MessagesUpsert#message_processable?` returns `false` if `jid_type != 'user'`. `jid_type` in `helpers.rb` returns `'lid'` for `@lid` JIDs. Result: the message is silently dropped — no logs beyond `Processing message X (fromMe: true)`, no contact/conversation/message created. Only human messages sent from inside the CRM appear; any external bot/IA outgoing is lost.

## Repro

1. Connect an Instagram/WhatsApp inbox that has an AI/automation sending replies via Evolution API directly
2. User sends inbound message → appears in CRM (jid has `@s.whatsapp.net`, OK)
3. External IA replies → webhook arrives with `remoteJid: "<numeric>@lid"` → DROPPED

## Root cause

```ruby
# app/services/whatsapp/evolution_handlers/messages_upsert.rb
def message_processable?
  return false if jid_type != 'user'   # <-- blocks 'lid'
  # ...
end

# app/services/whatsapp/evolution_handlers/helpers.rb
def jid_type
  case server
  when 's.whatsapp.net', 'c.us' then 'user'
  when 'g.us' then 'group'
  when 'lid' then 'lid'  # <-- not accepted downstream
  end
end
```

## Suggested fix

Resolve LID → phone via Evolution API's `POST /chat/fetchProfile/{instance}` with `{number: "<lid>@lid"}`. Response includes `wuid: "<phone>@s.whatsapp.net"` — use that for contact matching. Cache in Redis (we use 7d TTL).

Implementation sketch:

```ruby
def jid_type
  raw = @raw_message[:remoteJid] || @raw_message.dig(:key, :remoteJid)
  if raw.is_a?(String) && raw.end_with?('@lid')
    phone_jid = resolve_lid_via_evolution(raw.split('@').first)
    if phone_jid
      @raw_message[:key][:remoteJid] = phone_jid
      return 'user'
    end
  end
  # ... existing logic
end

[...]

PR #112, opened 2026-06-01
fix(whatsapp): resolve LID addressing on Evolution inbound (use remoteJidAlt)

firstplaceservdig1:
## Problem

WhatsApp's **LID addressing mode** (2024+ privacy feature) delivers 1:1 messages with `remoteJid: "<id>@lid"` instead of the phone JID. On the Evolution path:

- `jid_type` (`evolution_handlers/helpers.rb`) returns `'lid'` for `@lid`
- `message_processable?` only accepts `jid_type in %w[user group]`

→ the message is **silently dropped** (no contact, no conversation, no log beyond the event). In one production deployment this was **~99.6% of inbound 1:1 traffic** (1096 `@lid` vs 4 `@s.whatsapp.net` over 30 days). Related to #19 (which covers the outgoing side).

## Fix

Evolution already ships the real phone JID in **`remoteJidAlt`**. Add `effective_remote_jid` that prefers `remoteJidAlt` whenever the primary JID is a `@lid`, and use it from `jid_type` and `phone_number_from_jid`. Falls back to the raw `@lid` when `remoteJidAlt` is absent (no behavior change in that case).

```ruby
def effective_remote_jid
  jid = @raw_message[:remoteJid] || @raw_message.dig(:key, :remoteJid)
  return jid unless jid.to_s.end_with?('@lid')

  alt = @raw_message[:remoteJidAlt] || @raw_message.dig(:key, :remoteJidAlt)
  alt.presence || jid
end
```

## Verified in production

After deploying, 1:1 messages in LID mode now create a contact with the **real** phone number (from `remoteJidAlt`) plus a conversation — instead of being dropped.

## Tests

Adds `spec/services/whatsapp/incoming_message_evolution_service_lid_spec.rb`:
- LID + `remoteJidAlt` → `jid_type == 'user'`, phone extracted from `remoteJidAlt`, `message_processable?` true
- LID without `remoteJidAlt` → graceful fallback to `lid`
- normal `@s.whatsapp.net` → unchanged (regression guard)

## Refs

#19 (LID addressing — outgoing). This addresses the inbound 1:1 side.

## Summary by Sourcery

Handle WhatsApp Evolution inbound messages that use LID addressing by resolving the effective remote JID from remoteJidAlt so 1:1 messages are processed instead of dropped.

Bug Fixes:
- Ensure inbound 1:1 WhatsApp messages addressed via LID are recognized as user messages and no longer silently discarded.

Tests:
- Add specs covering LID inbound scenarios with and without remoteJidAlt and standard @s.whatsapp.net messages to guard the new behavior.

### DIFF
--- a/app/services/whatsapp/evolution_handlers/helpers.rb
+++ b/app/services/whatsapp/evolution_handlers/helpers.rb
@@ -112,7 +112,12 @@ module Whatsapp::EvolutionHandlers::Helpers
     return jid unless jid.to_s.end_with?('@lid')
 
     alt = @raw_message[:remoteJidAlt] || @raw_message.dig(:key, :remoteJidAlt)
-    alt.presence || jid
+    # Only trust remoteJidAlt when it is a real (non-LID) JID — guards against a
+    # future payload where the alternate is itself LID-style, which would otherwise
+    # be misclassified as 'user'. Falls back to the original LID when unavailable.
+    return alt if alt.present? && !alt.to_s.end_with?('@lid')
+
+    jid
   end
 
   def phone_number_from_jid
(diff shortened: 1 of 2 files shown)


firstplaceservdig1:
Thanks @sourcery-ai — both addressed in the latest commit:

1. **Single source of truth**: `group_jid` now also routes through `effective_remote_jid` (no behavior change today since group JIDs are `@g.us`, but keeps JID interpretation in one place).
2. **Defensive remoteJidAlt**: `effective_remote_jid` now only prefers `remoteJidAlt` when it is a real (non-`@lid`) JID — if a future payload has a LID-style alternate, we keep the original LID rather than misclassifying it as `user`. Added a spec for that case.

PR #115, opened 2026-06-05
feat(crm): unread conversation count endpoint + uncapped per-conversation count (EVO-1550)

marcelogorutuba:
## Summary
- Add `GET /api/v1/conversations/unread_count` returning the number of conversations with at least one unread incoming message for the current user, filtered by `Conversations::PermissionFilterService`.
- Replace the legacy `.last(10)` cap on `Conversation#unread_incoming_messages_count` with a real SQL `COUNT`, repointing the 4 call sites (`conversation_serializer`, `event_data_presenter`, `message`, `_conversation.json.jbuilder`).
- Replace the `LATERAL` join with `LIMIT 10` in `conversations_controller#unread_counts_map` with a direct subquery `COUNT(*)`.

## Security
- `unread_count` is scoped through `Conversations::PermissionFilterService`, so it respects inbox membership and user role.
- No new params accepted; endpoint is read-only.

## Test plan
- `bundle exec rspec spec/models/conversation_unread_count_spec.rb`
- `bundle exec rspec spec/controllers/api/v1/conversations/unread_count_spec.rb`
- `bundle exec rspec spec/serializers/conversation_serializer_spec.rb`

## Changed Files
- `app/models/conversation.rb`
- `app/serializers/conversation_serializer.rb`
- `app/presenters/event_data_presenter.rb`
- `app/models/message.rb`
- `app/views/api/v1/models/_conversation.json.jbuilder`
- `config/routes.rb`
- `spec/models/conversation_unread_count_spec.rb`
- `spec/controllers/api/v1/conversations/unread_count_spec.rb`
- `spec/serializers/conversation_serializer_spec.rb`

## Linked Issue
- EVO-1550

## Related PRs
- Frontend: will be linked once opened

## Summary by Sourcery

Add API support and internal plumbing for accurate unread conversation counts across the CRM.

New Features:
- Expose a new GET /api/v1/conversations/unread_count endpoint returning the number of conversations with unread incoming messages for the current user.

Enhancements:
- Compute per-conversation unread incoming message counts via SQL COUNT instead of a capped relation, and wire this into serializers, presenters, message events, and conversation JSON views.
- Optimize the unread counts map query to use a direct subquery-based COUNT instead of a LATERAL join with a hard LIMIT.

Tests:
- Add controller specs for the unread_count endpoint and model/serializer specs covering the new unread count behavior.


dpaes (changes requested):
## Code review — EVO-1550 backend (request changes)

Reviewed adversarially (multi-agent, reading the live branch) and cross-checked against the frontend PR #137. This PR is `CLEAN` on its own, but the feature is held: #137 is conflicting and the menu-aggregate semantics need a product decision. Backend findings below.

### ✅ Verified correct
- **Tenant isolation of `Conversation.all` is SAFE** — a schema invariant, not just an assumption: no `accounts` table, zero `account_id` columns, no `default_scope` on Conversation. The boundary is the same `PermissionFilterService` used by `index`/`filter` (admin→all, agent→`where(inbox: user.inboxes)`).
- Envelope contract is correct (`success_response(data:{unread_count})`, single `.data` wrap).
- The 4 call-site repoints are safe (the old `unread_incoming_messages` returned an Array `.last(10)`; every caller only used `.count`, so the scalar swap is 1:1). The contact read-receipt path now reconciles to the true (uncapped) count — an AC1 fix.

### 🔴 High
- **AC5/AC6/AC7 unit mismatch.** The endpoint returns `.distinct.count('conversations.id')` (`conversations_controller.rb:241`) = number of **conversations** with ≥1 unread, while the list badge counts **messages**. Crisp AC6 violation: a 2nd unread message in an already-unread conversation does not change the distinct count, so the menu doesn't increment; AC7 ("decrement by the conversation's value") can only drop by 1, never by N. Decide with the PM: SUM messages in the endpoint, or relabel the menu as "unread conversations" and adjust AC5/6/7.
- **Admin path is uncapped & table-wide.** `PermissionFilterService` returns `Conversation.all` unfiltered for administrators (the common role) → the aggregate scans the whole table with no `LIMIT`/cap. The card's "Plano B: cap 999" is not implemented. Combined with the frontend calling this endpoint on every websocket message (see #137), a busy admin account hammers an unbounded aggregate. Add a cap, or bound the admin query, or wire local deltas + debounce on the frontend.

### 🟡 Medium
- **Read-flag divergence.** Only the list SQL (`unread_counts_map`) filters `content_attributes->>'read' IS DISTINCT FROM 'true'` (`:408`); the new endpoint and `unread_incoming_messages_count` (`conversation.rb:180-186`) do not. Inert today (nothing writes `read=true` in this repo), but a real maintenance trap — pick one canonical predicate for all three counts.
- **N+1 made heavier.** `available_for_pipeline` (`:226`) and `pipeline_items_controller` serialize without passing `unread_counts:`, so each row runs an uncapped `COUNT` via the model method (was a `LIMIT 10` fetch before). Pass the batched `unread_counts_map`.
- **Specs are mock-only.** `unread_count_spec.rb` / `conversation_unread_count_spec.rb` stub the whole AR chain (`.count.and_return(17/7/42)`) — zero real SQL, zero rows, zero of the isolation boundary the safety argument depends on. They are also **not in the CI staleness-guard allowlist**, so they never run in CI. Add at least one request/integration spec creating real rows (N>10, before/after `agent_last_seen_at`) and run `bundle exec rspec` locally.

### ⚪ Low/nit
- Model method inlines the count instead of reusing the `unread_messages.incoming.count` scope the card asked for (functionally equivalent — optional refactor for a single source of truth).
- The card-required composite index `(conversation_id, message_type, created_at)` wasn't added, but develop already ships a stronger partial index `idx_messages_conv_created_incoming_desc` `(conversation_id, created_at) WHERE message_type=0` that covers the predicate. Recommend an `EXPLAIN ANALYZE` on a large account to confirm the planner uses it for the `COALESCE(... to_timestamp(0))` range.
- Legacy `unread_incoming_messages` (`conversation.rb:176-178`) is now fully orphaned (zero callers) — safe to delete.
- The `LoadError` scaffold + top-level `return unless defined?(Rails)` in `unread_count_spec.rb` can silently no-op; the sibling model spec uses the plain `require 'rails_helper'` — prefer that.

Holding for changes (H2/H3) and pending the rebase of the paired frontend PR #137.

dpaes (commented):
## Re-review — EVO-1550 (backend) · 💬 Comment (correct; one optional follow-up)

The backend is correct and ships. Reviewed as part of the EVO-1550 re-review; the light request-change for this feature lands on the frontend (#137, fetch coalesce), so this is a comment, not a block.

**Verified ✅**
- The new `GET /conversations/unread_count` returns `COUNT(DISTINCT conversations.id)` (conversations unit, matches the rewritten AC5), scoped by `PermissionFilterService`.
- Account-safety confirmed: single-account by design (no `Account` model / `account_id` column); `Conversation.all` is the whole single-account table, boundary is `inbox: user.inboxes`. The earlier cross-account concern stays retracted.
- The 4 repointed call-sites all consume the new Integer method bare (no `.count`/`.map`/`.last` on it) — no breakage from the Array→Integer return change. The old `.last(10)` method now has zero callers.

**H3 (cap) — optional, your call**
The uncapped `COUNT(DISTINCT)` is predicate-covered by the pre-existing partial index `idx_messages_conv_created_incoming_desc` (so no seq-scan), but it doesn't bound cardinality: the admin path (`Conversation.all`) is an indexed table-wide count on a polled endpoint. Mitigated, not eliminated. Once the frontend fetch is coalesced (#137), the polling pressure drops a lot; a cap-999 (the documented Plan B) is optional belt-and-suspenders for large/admin-heavy deployments.

**Low**
- Read-flag divergence (`content_attributes->>'read'` filtered in `unread_counts_map` but not in the new method/endpoint) is **inert** — nothing in the repo writes that flag, so every path reduces to `created_at > agent_last_seen_at` today. Latent only if a writer is ever added.
- The `_conversation.json.jbuilder` loaded-association branch counts by the (never-written) read flag and ignores `agent_last_seen_at` — pre-existing, on a line this PR touches; worth unifying for consistency.
- Specs are mock-only and don't exercise the real SQL; crm-community has no RSpec in CI.

dpaes (approved):
## Re-review — Approved ✅ (backend half of EVO-1550)

Backend is unchanged since my prior review and re-confirmed clean:
- `GET /conversations/unread_count` uses `COUNT(DISTINCT conversations.id)` scoped through `PermissionFilterService` (unit = conversations, matching the menu semantics).
- `success_response` envelope is single; the frontend `extractData` unwraps it so the number arrives intact (no double-envelope / NaN).
- Single-account safety holds (no `Account` model; boundary is `user.inboxes`).

Branch is behind `develop` (no conflict — clean three-dot diff). The Round-5 work that closed the blocker was frontend-only (see #137).

**CI caveat:** crm-community runs only Sourcery (no RSpec gate); the `unread_count` specs are mock-chain (prove the chain is invoked, not the SQL). Worth a local `bundle exec rspec`.

Merging to `develop`.

PR #116, opened 2026-06-05
feat(crm): email_team + canned_response show endpoints for the evo-flow journey executor (EVO-1634)

nickoliveira23:
## Summary
Adds the two CRM endpoints the evo-flow Journey (Temporal) executor needs so Flow Builder action nodes execute their real effect — part of **EVO-1634** (wire palette action nodes into the evo-flow runtime).

- `GET /api/v1/canned_responses/:id` (**show**) — lets `send-canned-response-node` resolve content by id with a single fetch (replaces the previous list-and-scan).
- `POST /api/v1/conversations/:id/email_team` — backs `send-email-team-node`; validates `team_ids` + `message` and notifies each team via `TeamNotifications::AutomationNotificationMailer.conversation_creation(...).deliver_later`.

Both are service-token authenticated (`X-Service-Token` = `EVOAI_CRM_API_TOKEN`), matching the existing evo-flow → CRM calls.

## Validation
- `bundle exec rspec spec/requests/api/v1/conversations_email_team_spec.rb spec/requests/api/v1/canned_responses_show_spec.rb` → 3 examples, 0 failures
- `ruby -c` on the changed controllers

## Changed Files
- `app/controllers/api/v1/canned_responses_controller.rb`
- `app/controllers/api/v1/conversations_controller.rb`
- `config/routes.rb`
- `spec/requests/api/v1/canned_responses_show_spec.rb`
- `spec/requests/api/v1/conversations_email_team_spec.rb`

## Related PRs
- evo-flow#25 — executor node wiring (consumes these endpoints)
- evo-ai-frontend-community — palette parity manifest (link added after open)

## Linked Issue
- EVO-1634

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Add new CRM API endpoints to support evo-flow journey executor actions for emailing teams and fetching canned responses by ID.

New Features:
- Introduce POST /api/v1/conversations/:id/email_team to trigger team notification emails for a conversation.
- Expose GET /api/v1/canned_responses/:id to retrieve a single canned response by ID.

Tests:
- Add request specs covering the canned response show endpoint and the conversations email_team endpoint.

PR #117, opened 2026-06-05
fix(automation): execute and record contact-triggered automations (EVO-1635)

daniloleonecarneiro:
## EVO-1635 — Automações de contato não executam nem aparecem nos logs

**Root cause:** `Contact#dispatch_*_event` desabilitado → eventos de contato nunca chegavam ao `AutomationRuleListener`; e o ramo de contato não gravava run (sem `RunRecorder`) e só rodava webhook.

**Fix:** `AutomationContactEventJob` (ponte só p/ automação), `ContactActionService` (execução nativa de contato + skip observável), callbacks no Contact com guarda de loop, `RunRecorder` no ramo de contato.

**Testes:** 16 specs verdes (service + listener + job + model).

Parte da EVO-1637. Base da stack (1635 → 1640 → 1638).

### DIFF
--- a/app/jobs/automation_contact_event_job.rb
+++ b/app/jobs/automation_contact_event_job.rb
@@ -11,7 +11,13 @@
 # and double the outbound webhooks already sent via EvoFlow. Instead this job
 # invokes the automation listener directly, in the background, for contacts only.
 class AutomationContactEventJob < ApplicationJob
-  queue_as :critical
+  # Contact create/update is far higher-volume than conversation events (bulk
+  # imports/syncs write thousands of contacts), so this runs on the normal queue
+  # rather than :critical — automation is best-effort background work and must
+  # not starve realtime/critical conversation handling during a mass import.
+  # Contacts with no matching active rule are filtered upstream
+  # (Contact#enqueue_contact_automation) so they never reach this queue at all.
+  queue_as :default
 
   SUPPORTED_EVENTS = %w[contact_created contact_updated].freeze
 
--- a/app/jobs/automation_contact_event_job.rb
+++ b/app/jobs/automation_contact_event_job.rb
@@ -1,0 +1,27 @@
+# Feeds contact_created / contact_updated events to the automation engine ONLY.
+#
+# Contact#dispatch_create_event / #dispatch_update_event (the full sync+async
+# dispatcher path) are intentionally disabled — contacts publish through Wisper
+# (EvoFlow) for integrations. That left AutomationRuleListener, which subscribes
+# to the AsyncDispatcher, with no source of contact events, so contact-triggered
+# automation rules never ran.
+#
+# We deliberately do NOT re-enable the broad dispatch: that would also resurrect
+# the legacy HookListener / WebhookListener / ActionCableListener for contacts
+# and double the outbound webhooks already sent via EvoFlow. Instead this job
+# invokes the automation listener directly, in the background, for contacts only.
+class AutomationContactEventJob < ApplicationJob
+  queue_as :critical
+
+  SUPPORTED_EVENTS = %w[contact_created contact_updated].freeze
+
+  def perform(event_name, contact_id, changed_attributes = {})
+    return unless SUPPORTED_EVENTS.include?(event_name)
+
+    contact = Contact.find_by(id: contact_id)
+    return if contact.nil?
+
+    event = Events::Base.new(event_name, Time.zone.now, { contact: contact, changed_attributes: changed_attributes })
+    AutomationRuleListener.instance.public_send(event_name, event)
+  end
+end
--- a/app/listeners/automation_rule_listener.rb
+++ b/app/listeners/automation_rule_listener.rb
@@ -90,18 +90,11 @@ class AutomationRuleListener < BaseListener
     rules = current_account_rules('contact_created', account)
 
     rules.each do |rule|
-      # Para eventos de contato que só têm condições de contato, 
-      # não precisamos de uma conversa
+      # Para eventos de contato que só têm condições de contato (ou nenhuma),
+      # não precisamos de uma conversa. Avalia + executa via execução nativa de
+      # contato, registrando o run no automation_rule_runs (observabilidade).
       if rule_has_only_contact_conditions?(rule)
-        conditions_match = evaluate_contact_conditions(rule, contact, changed_attributes)
-        if conditions_match
-          # Executa ações que não precisam de conversa (como webhooks)
-          if rule.mode == 'flow' && rule.flow_data.present?
-            AutomationRules::FlowExecutionService.new(rule, account, nil, contact).perform
-          else
-            execute_contact_actions(rule, account, contact)
-          end
-        end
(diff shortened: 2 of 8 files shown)


dpaes (approved):
## Review — EVO-1635 · ✅ Approved (code) — merge gated on stack rebase

Reviewed against the EVO-1637 acceptance criteria. Solid, well-reasoned foundation for the contact-automation engine.

**What's solid**
- The narrow `AutomationContactEventJob` bridge feeds only `AutomationRuleListener` and deliberately avoids re-enabling the broad dispatcher (no double webhooks via the legacy Hook/Webhook/ActionCable listeners). Good call, well-documented.
- Loop guard via `Current.executed_by` holds: `ContactActionService#perform` sets it before the synchronous `update!`, whose `after_update_commit` re-enqueue check runs in the same thread → guarded. The flow branch is also covered (FlowExecutionService sets `Current.executed_by` in its initializer).
- `RunRecorder` on the contact path makes runs visible in the logs (matched / no_match / error); conversation-bound actions are recorded as `skipped` with a reason instead of silently no-op'ing.
- `add_label`/`remove_label` correctly route through the `update!(label_list:)` **setter** so dirty-tracking fires (`saved_change_to_label_list?` / `publish_label_changes`) — the right pattern for acts-as-taggable-on.

**Non-blocking (recommend, not gating)**
- ⚠️ Queue choice: `queue_as :critical` fires on **every** contact create/update. Bulk imports/syncs will flood the critical queue (the spam guard throttles *execution*, not *enqueues*). Recommend a dedicated/normal queue, and optionally `AutomationRule.exists?(event_name:, active: true)` before enqueuing so contacts with no rules don't spawn an empty job each write.
- Account scope (systemic, **pre-existing** — not introduced here): `current_account_rules` ignores its `account` arg and returns rules across all accounts; every listener path passes `account=nil`. The contact path — previously a silent no-op — now *executes*, so this account-blindness becomes live for contacts. If the community edition is multi-account, an account B `contact_updated` rule would evaluate/act on an account A contact. Worth a conscious decision; out of scope for this PR.

Caveat: crm-community runs no RSpec in CI, so the green-specs claim is self-reported. Recommend `bundle exec rspec` locally, plus a seam test that drives a real `update!(label_list:)` through the enqueue and asserts the payload (the specs currently hand-inject the dispatch hash, so the producer→consumer label path isn't exercised).

Note (stack): approving the **code**. Do **not** squash-merge this before the stack is rebased — merging #117 will auto-close #118/#119 (base disappears). Will coordinate the rebase with @Danilo.

daniloleonecarneiro:
Rebased onto `develop` (now includes #116 / EVO-1634) — zero conflicts, disjoint surfaces. Force-pushed `970776c` → `bbe9a85`.

Review follow-up addressed in `bbe9a85`:
- Enqueue guard: skip `AutomationContactEventJob` when no active rule listens (`AutomationRule.exists?`) — no more empty job per contact write.
- Job queue `:critical` → `:default` (contact writes are bulk-volume; keep them off the realtime/critical path).
- Specs: active rule on happy-path, no-rule-no-enqueue guard test, and a seam test driving a real `update!(label_list:)` through the producer path.

Local `rspec` is green (18/18) only against an **isolated** DB — the dockerized container sets `POSTGRES_DATABASE=evo_community`, so `RAILS_ENV=test` otherwise runs against the dev DB and leftover rules cause false `count changed by 2` failures. Ran with `-e POSTGRES_DATABASE=evolution_test`.

Account-scope note (`current_account_rules`): `automation_rules` has no `account_id` column — community is account-agnostic at the rules layer by design, so this is a conscious won't-fix here (would need a schema migration; deserves its own card if multi-account becomes a goal).

dpaes (approved):
## Re-review — EVO-1635 · ✅ Approved (code) — merging

Re-reviewed against the EVO-1637 acceptance criteria after the rebase + review-feedback round. Both prior points are resolved and the rebase is clean (`bbe9a85` sits directly on develop tip `3f59646`; base `develop`, MERGEABLE/CLEAN).

### Prior point 1 — `:critical` queue + empty job → ✅ resolved (`bbe9a85`)
- `queue_as :default` — contact writes are far higher-volume than conversation events; automation is best-effort and must not starve the critical/realtime queue during a mass import.
- `Contact#enqueue_contact_automation` now early-returns unless `AutomationRule.exists?(event_name:, active: true)` — contacts with no active rule never enqueue an empty job.
- Added the seam test that drives a real `update!(label_list:)` through the producer path (exercises the acts-as-taggable-on setter → `after_update_commit` → enqueue), plus a no-rule-no-enqueue guard test.

### Prior point 2 — account-scope blindness → ✅ deferral is correct
Confirmed in `db/schema.rb`: `automation_rules` has **no `account_id`** column (and neither does `automation_rule_runs`). The table is global-to-the-deployment by design in community — `current_account_rules` can't scope by an account dimension that doesn't exist. A real fix needs a migration + re-scoping every listener path, well beyond a contact-trigger bugfix. Correctly recorded as a conscious decision; worth its own card if multi-account ever becomes a community goal.

### Adversarial verification
Ran an independent multi-agent verification pass (runtime/control-flow, data/contract, regression/scope, plus a completeness critic) against the post-rebase tree. All converged on **no blockers**. Spot-confirmed: single-feed-path loop guard (`Current.executed_by` set by `ContactActionService` before the synchronous `after_update_commit`), no Wisper double-event (listener is fed only by the job, not the dispatcher), `Events::Base` signature match, `previous_changes` not clobbered by the `publish_*` callbacks, `add_label`/`remove_label` param contract = label IDs (matches `FlowExecutionService` + the conditions evaluator), and `automation_rule_runs` schema + status validation match `RunRecorder` exactly.

### Non-blocking (no follow-up cards unless prioritized)
- Two **dangling orphan comments** in `automation_rule_listener.rb` (≈L206-209 and ≈L441-444) describe run-recording for the spam-guard and the conversation-condition branch that isn't implemented here — that's EVO-1640. Remove them or turn them into explicit EVO-1640 TODOs so they don't read as done.
- `ContactActionService#add_label`/`#remove_label`: a rule referencing a **deleted** label resolves to an empty set and records the step as `success` (mirrors conversation-side `ActionService`); a precision nit, not silent (the run is still logged).
- `RunRecorder#persist!` rescues `StandardError` → a genuinely failing `create!` would drop the run with only a log line. Fine today (validations are satisfied), worth keeping in mind for observability.
- `contact_created` lacks the empty-`changed_attributes` guard that `contact_updated` has — harmless (`previous_changes` on create is never empty), pure consistency nit.
- No composite index on `automation_rules(event_name, active)` backing the new `exists?` guard — negligible on a tiny table, deferred perf.

### Verdict
**Approved.** Squash-merging to develop.

⚠️ **Stack:** #118 (EVO-1640) is based on this branch and is already CONFLICTING — it needs a rebase onto develop after this lands (then #119/EVO-1638). The `delete_branch_on_merge` repo setting is off and I'm merging **without** `--delete-branch`, so #117's branch is kept and #118's base survives until it's rebased.

PR #118, opened 2026-06-05
fix(automation): record skipped runs for contact silent drops (EVO-1640)

daniloleonecarneiro:
## EVO-1640 — Observabilidade: drops silenciosos

**Fix:** spam guard de `contact_updated` agora ENV-configurável e grava `skipped` (uma vez por janela); ramo de condição-de-conversa sem conversa grava `skipped`/`no_match`/`matched` (antes era drop silencioso).

**Testes:** 17 specs verdes (base + observability T2).

Parte da EVO-1637. **Stacked sobre** `danilocarneiro/evo-1635-bug-cadastro-de-automacao` (EVO-1635) — merge sequencial.

PR #119, opened 2026-06-05
fix(automation): contact condition operators + labels semantics (EVO-1638)

daniloleonecarneiro:
## EVO-1638 — Condições de contato: never-match silencioso + divergência de labels

**Fix:** `evaluate_contact_conditions` ganha `attribute_changed` (scalar + labels) e `starts_with`; `labels equal_to` alinhado a any-of (era all-of vs SQL).

**Testes:** 22 specs verdes (tudo).

Parte da EVO-1637. **Stacked sobre** `danilocarneiro/evo-1640-bug-automacao-drops-silenciosos-na-observabilidade-spam` (EVO-1640) — merge sequencial.

### DIFF
--- a/app/listeners/automation_rule_listener.rb
+++ b/app/listeners/automation_rule_listener.rb
@@ -382,6 +382,14 @@ class AutomationRuleListener < BaseListener
   # in previous_changes ([[old_titles], [new_titles]]); scalars under their column.
   def contact_attribute_changed_match?(attribute_key, values, changed_attributes)
     changed = (changed_attributes || {}).with_indifferent_access
+    # `attribute_changed` needs a {from, to} transition shape. `nil` means
+    # "changed at all" (wildcard); a malformed value (e.g. a bare Array from a
+    # legacy/broken condition) can't express a transition. Treat it as no-match
+    # for THIS condition instead of letting `values['from']` raise a TypeError —
+    # that exception is rescued upstream but errors the ENTIRE rule run rather
+    # than failing the single condition.
+    return false unless values.nil? || values.is_a?(Hash)
+
     values ||= {}
     backend_key = attribute_key == 'labels' ? 'label_list' : attribute_key
     transition = changed[backend_key]
--- a/app/listeners/automation_rule_listener.rb
+++ b/app/listeners/automation_rule_listener.rb
@@ -318,83 +318,102 @@ class AutomationRuleListener < BaseListener
   end
 
   def evaluate_contact_conditions(rule, contact, changed_attributes)
-    # Avalia condições simples de contato
+    # Avalia condições de contato sem precisar de conversa.
     rule.conditions.all? do |condition|
       attribute_key = condition['attribute_key']
       filter_operator = condition['filter_operator']
       values = condition['values']
 
-      case attribute_key
-      when 'labels'
-        # Para labels, verifica se o contato tem as labels especificadas
-        contact_labels = contact.label_list
-        case filter_operator
-        when 'equal_to'
-          label_ids = values
-          label_titles = Label.where(id: label_ids).pluck(:title)
-          (label_titles - contact_labels).empty?
-        when 'not_equal_to'
-          label_ids = values
-          label_titles = Label.where(id: label_ids).pluck(:title)
-          (label_titles & contact_labels).empty?
-        when 'is_present'
-          contact_labels.any?
-        when 'is_not_present'
-          contact_labels.empty?
-        else
-          false
-        end
-      when 'name', 'email', 'phone_number', 'identifier'
-        # Atributos simples do contato
-        contact_value = contact.send(attribute_key)
-        case filter_operator
-        when 'equal_to'
-          values.include?(contact_value)
-        when 'not_equal_to'
-          !values.include?(contact_value)
-        when 'contains'
-          values.any? { |v| contact_value&.include?(v) }
-        when 'does_not_contain'
-          values.none? { |v| contact_value&.include?(v) }
-        when 'is_present'
-          contact_value.present?
-        when 'is_not_present'
-          contact_value.blank?
-        else
-          false
-        end
-      when 'blocked'
-        case filter_operator
-        when 'equal_to'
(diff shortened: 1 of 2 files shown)


dpaes (approved):
## Review — EVO-1638 · ✅ Approved (code) — merge gated on stack rebase

Reviewed against the EVO-1637 acceptance criteria. The operator additions are correct and, where checkable, align with the conversation/SQL path.

**Verified (not just claimed)**
- `labels equal_to` all-of → any-of is the **right** change: `ConditionsFilterService` uses `EXISTS ... name IN (:value)` (any-of), so `(label_titles & contact_labels).any?` aligns the contact path to the SQL path. The transition-wildcard semantics (empty `from`/`to` = wildcard; added/removed intersection) also match `scalar_transition_match?` / `labels_transition_match?`.
- The `labels` + `attribute_changed` path **does fire** end-to-end. I chased a suspected "the job payload omits `label_list`" blocker to ground: acts-as-taggable-on 12.0.0 registers `label_list` as a real ActiveModel attribute (`core.rb:43`) and the setter calls `write_attribute` (`core.rb:58`), so `update!(label_list:)` populates `previous_changes['label_list'] = [[old],[new]]` (TagList < Array, no `as_json` override) — exactly what `contact_labels_transition_match?` expects. Confirmed; not a blocker.

**Non-blocking (recommend, not gating)**
- `city` / `company` + `attribute_changed` would silently never-match: only `labels` is aliased to a backend key, but `city`/`company` live in the `additional_attributes` jsonb (changes land under `previous_changes['additional_attributes']`, not `['city']`). Latent — `filter_keys.yml` doesn't expose `attribute_changed` for these, and it's the same limitation as the SQL path, so not a regression. If `attribute_changed` is ever exposed for them, alias `city/company/country_code → additional_attributes`. (Note `country_code` IS a real column → producer/consumer asymmetry vs the SQL path, which treats it as additional_attributes.)
- The "mirrors ConditionsFilterService exactly" note is slightly off — and in the PR's favor: the `.to_s` coercion here makes `blocked` match correctly (`previous_changes['blocked'] = [false, true]` booleans vs `['false','true']` string values), whereas the SQL `scalar_transition_match?` compares raw and likely has a **latent boolean never-match** on `blocked`. Worth a follow-up on the SQL side.
- `attribute_changed` with a malformed `values` (an Array instead of `{from,to}`) raises `TypeError` on `values['from']`; it's caught by the `rescue` in `evaluate_and_execute_contact_rule`, but that aborts the **entire** rule as an `error` run (`.all?` over all conditions), not a clean per-condition `no_match`. Consider guarding the shape.

Caveat: the producer→consumer label seam is confirmed by code/gem analysis, not by the suite — the specs hand-inject `{'label_list' => [[],['vip']]}` rather than driving `update!(label_list:)`. Recommend a real seam test (and `bundle exec rspec`, since crm-community has no RSpec in CI).

Note (stack): approving the **code**. Top of the stack (on #118 on #117); merges to `develop` only after the stack is rebased. Coordinating with @Danilo.

daniloleonecarneiro:
Rebased onto `develop` after #118 merged (squash `0d63796`).

- `git rebase --onto origin/develop 212dfca` → dropped the old 1635+1640 commits (SHAs differ from the squashes), replayed only the 1638 delta. `0a12673` → `1c2d7a4`, force-with-lease.
- Retargeted this PR base from the (merged) 1640 branch → `develop`. Now MERGEABLE.

Review follow-up addressed in `1c2d7a4`:
- **Seam test (your ressalva):** added a spec that drives a real `update!(label_list: ['vip'])` and feeds the listener the actual `contact.previous_changes` (asserting it carries `label_list`), instead of hand-injecting the transition hash — confirms the producer→consumer seam empirically.
- **Guard the shape:** `contact_attribute_changed_match?` now treats a malformed `values` (bare Array vs `{from,to}`) as a per-condition `no_match` instead of raising a `TypeError` that the rescue turns into a whole-rule `error` run. +spec.

Deliberately **not** addressed (documented on Linear): `city`/`company` + `attribute_changed` latent never-match — same jsonb `additional_attributes` limitation as the SQL path, not a regression; deserves its own engine card alongside the `blocked` SQL asymmetry you flagged.

Validation: **26 examples, 0 failures** against an isolated DB (`-e POSTGRES_DATABASE=evolution_test`).

dpaes (approved):
## Re-review — EVO-1638 · ✅ Approved — merging

Post-rebase re-review (base now `develop`, MERGEABLE/CLEAN; contract + staleness checks green). The isolated 1638 delta (commit `8976f46`) is byte-identical to what was approved: `labels equal_to` all-of→any-of (aligned with the SQL `EXISTS … IN`), `attribute_changed` (scalar + labels transition mirroring `scalar/labels_transition_match?`, reading labels from `previous_changes['label_list']`), `starts_with`, helpers + specs.

### New commit `1c2d7a4` — resolves both review nits ✅
- **Malformed-shape guard**: `return false unless values.nil? || values.is_a?(Hash)` in `contact_attribute_changed_match?` → a malformed `values` (bare Array from a legacy/broken condition) now fails *that condition* as `no_match` instead of raising a `TypeError` that errored the entire rule run. `nil` still means "changed at all" (wildcard). Correct.
- **Real seam test**: drives `contact.update!(label_list: ['vip'])` and feeds `contact.previous_changes.as_json` to the matcher, asserting `has_key('label_list')` + a `matched` run — empirically validating the acts-as-taggable-on producer→consumer seam (previously only confirmed by reading the gem). Plus a shape-guard test (malformed values → `no_match`, not `error`).

### Non-blocking (unchanged, latent)
- `city`/`company` + `attribute_changed`: latent silent never-match (they live in the `additional_attributes` jsonb, so the change lands under `previous_changes['additional_attributes']`, not the bare key) — same limitation as the SQL path, not a regression.

Caveat: crm-community doesn't run RSpec in CI → specs are self-reported (dev ran the contact suite locally).

### Verdict
**Approved.** Squash-merging to `develop`. Top of the stack — no child PR — so after merge the `evo-1635`/`evo-1640`/`evo-1638` branches are all free to delete (verified).

PR #123, opened 2026-06-08
feat(crm): SendGrid email channel — model, migration, CRUD (EVO-1248)

nickoliveira23:
## Summary
First story of the SendGrid Email Channel front (epic E). Adds `Channel::Sendgrid` as a new polymorphic channel type alongside the existing channels, wired into the existing inbox channel CRUD.

- New table `channel_sendgrid` (`api_key_encrypted`, `from_email`, `from_name`, `reply_to`, `sender_domain`).
- New model `Channel::Sendgrid` with the API key **encrypted at rest** (Fernet, reusing `InstallationConfig.encryption_key`) and exposed via a virtual `api_key` accessor — `api_key_encrypted` is never accepted as a param nor serialized.
- Wired into the channel dispatch (`account_channels_method` + `channel_type_from_params`) and the `create_channel` whitelist.
- `InboxSerializer` exposes only safe fields (`from_email`, `from_name`, `reply_to`, `sender_domain`, `api_key_present`) and excludes `api_key_encrypted` even on `include_channel: true`.
- Validations: `from_email` (presence + format), `reply_to` (format, optional), `sender_domain` (format, optional), `api_key` (presence).

Single-tenant: no `account_id`/`tenant_id` copied from the BMS reference.

## Security
- API key stored encrypted (Fernet token, `gAAAAA…`); plaintext never persisted.
- Strong params accept `api_key` (virtual) only — never `api_key_encrypted`.
- Serializer never emits `api_key`/`api_key_encrypted`; only a boolean `api_key_present`.
- Verified end-to-end: create response and full-channel serialization contain no plaintext key.

## Reinterpreted AC (accepted)
The card's AC #3/#5 reference `POST/GET /api/v1/notificame/channels`, but that route does **not** perform channel CRUD in this codebase — channel CRUD is the polymorphic `/api/v1/inboxes` flow (same as Gmail/Outlook/WhatsApp). Implemented there; the AC behavior (201 + no plaintext key; appears in list alongside email channels) is satisfied on the real endpoint. Note: "Gmail/Outlook" are `provider` values on `Channel::Email`, not separate channel types — SendGrid is a dedicated type per the card's explicit "criar model SendGridChannel".

## Test plan
- `bundle exec rspec spec/models/channel/sendgrid_spec.rb spec/requests/api/v1/inboxes_sendgrid_spec.rb spec/serializers/inbox_serializer_sendgrid_spec.rb` → 16 examples, 0 failures
- `bundle exec rubocop` on changed files → clean (one pre-existing offense in inboxes_helper unrelated to this change)
- `bin/rails db:migrate`
- Live smoke (curl against a running server): POST valid → 201 (no plaintext key); POST invalid `from_email` → 422; GET list shows `Channel::Sendgrid`; DB confirms `api_key_encrypted` is a Fernet token.

## Changed Files
- `db/migrate/20260608194533_create_channel_sendgrid.rb` (new)
- `db/schema.rb`
- `app/models/channel/sendgrid.rb` (new)
- `app/models/inbox.rb` (`sendgrid?` predicate)
- `app/helpers/api/v1/inboxes_helper.rb` (dispatch)
- `app/controllers/api/v1/inboxes_controller.rb` (dispatch + create whitelist)
- `app/serializers/inbox_serializer.rb` (safe fields + secret exclusion)
- `spec/models/channel/sendgrid_spec.rb` (new)
- `spec/requests/api/v1/inboxes_sendgrid_spec.rb` (new)
- `spec/serializers/inbox_serializer_sendgrid_spec.rb` (new)

## Linked Issue
- EVO-1248

### DIFF
--- a/app/controllers/api/v1/inboxes_controller.rb
+++ b/app/controllers/api/v1/inboxes_controller.rb
@@ -763,7 +763,7 @@ module Api
         end
 
         def create_channel
-          return unless %w[web_widget api email line telegram whatsapp sms].include?(permitted_params[:channel][:type])
+          return unless %w[web_widget api email line telegram whatsapp sms sendgrid].include?(permitted_params[:channel][:type])
 
           # Debug logs for Evolution Go channel creation
           if permitted_params[:channel][:type] == 'whatsapp' && permitted_params[:channel][:provider] == 'evolution_go'
@@ -900,7 +900,8 @@ module Api
             'line' => Channel::Line,
             'telegram' => Channel::Telegram,
             'whatsapp' => Channel::Whatsapp,
-            'sms' => Channel::Sms
+            'sms' => Channel::Sms,
+            'sendgrid' => Channel::Sendgrid
           }[permitted_params[:channel][:type]]
         end
 
--- a/app/helpers/api/v1/inboxes_helper.rb
+++ b/app/helpers/api/v1/inboxes_helper.rb
@@ -1,4 +1,4 @@
-module Api::V1::InboxesHelper
+module Api::V1::InboxesHelper # rubocop:disable Metrics/ModuleLength
   def inbox_name(channel)
     if channel.is_a?(Channel::Telegram)
       # Usar o bot_name do canal (obtido da API do Telegram via first_name)
@@ -114,7 +114,8 @@ module Api::V1::InboxesHelper
       'line' => Channel::Line,
       'telegram' => Channel::Telegram,
       'whatsapp' => Channel::Whatsapp,
-      'sms' => Channel::Sms
+      'sms' => Channel::Sms,
+      'sendgrid' => Channel::Sendgrid
     }[permitted_params[:channel][:type]]
   end
 
--- a/app/models/channel/sendgrid.rb
+++ b/app/models/channel/sendgrid.rb
@@ -1,0 +1,62 @@
+# == Schema Information
+#
+# Table name: channel_sendgrid
+#
+#  id                :uuid             not null, primary key
+#  api_key_encrypted :text             not null
+#  from_email        :string           not null
+#  from_name         :string
+#  reply_to          :string
+#  sender_domain     :string
+#  created_at        :datetime         not null
+#  updated_at        :datetime         not null
+#
+# Indexes
+#
+#  index_channel_sendgrid_on_from_email  (from_email)
+#
+
+class Channel::Sendgrid < ApplicationRecord
+  include Channelable
+
+  self.table_name = 'channel_sendgrid'
+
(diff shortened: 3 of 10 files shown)


dpaes (approved):
## Review — Approved ✅

Reviewed the highest-risk surface (API-key encryption + secret exposure) by reading the actual code across three independent lenses, not the QA notes.

**Security / no-leak — clean:**
- **Encryption at rest is real.** `api_key` is a virtual method: the setter encrypts via Fernet (`InstallationConfig.encryption_key`) into the `api_key_encrypted` column — the only column on the table (no plaintext column in the migration/schema). Plaintext is never persisted and never held on the instance (only a transient local in the setter). Matches the established `InstallationConfig` convention and fails **loud** on missing key.
- **No serializer leak.** Because `api_key` is a method (not a column), `as_json` physically cannot serialize it; `api_key_encrypted` is explicitly excluded in the `include_channel: true` branch via the `respond_to?` guard. The serializer spec asserts absence of both keys **and** that the full payload doesn't contain the plaintext.
- **Strong params** permit only the virtual `:api_key`, never `:api_key_encrypted` — mass-assignment of the ciphertext column is impossible. Bonus: `filter_parameters` (`:_key`) scrubs the inbound key from request logs.
- Decrypt fails **closed** (`rescue → nil`, logs error); key rotation overwrites the ciphertext (old cipher dropped).

**Acceptance criteria — all 5 MET** (schema fields, encrypt-at-rest, 201 without plaintext, 422 on invalid `from_email` via the controller's `RecordInvalid` rescue, list-alongside Gmail/Outlook).

**Endpoint reinterpretation — justified.** The card cites `/api/v1/notificame/channels`, but that controller is a WhatsApp-BSP proxy with only an `index` action (no channel CRUD). The canonical channel-CRUD path is `/api/v1/inboxes`, which every other channel type uses; SendGrid is wired into all three type maps + `inbox.rb` + the serializer. The AC behaviors hold on the implemented endpoint.

**Follow-up (LOW, non-blocking — merging as-is):** the schema.rb regen also rewrote two unrelated `products` `check_constraint` lines to a different (semantically equivalent) normalized form — local-Postgres dump noise. Not a functional change. Heads-up sent to align the local PG version with CI/develop to avoid future `schema.rb` churn.

**CI caveat:** crm-community does not run RSpec in CI (Sourcery only), so the "16 examples, 0 failures" is author-self-reported. The specs are genuinely strong (encryption-at-rest inspected, full-body leak scan, real 422 path) and would prove the behavior if green — recommend a local `bundle exec rspec` before relying on them.

Merging to `develop`.

PR #128, opened 2026-06-09
feat(api): server-side mask of contact PII for non-admin users (EVO-1551)

marcelogorutuba:
## Summary
- New `ContactPiiMasker` helper (`app/lib/`) — Ruby port of the frontend masking rules
- Masks `phone_number`, `email`, `identifier`, `source_id` and phone-shaped `name` server-side when account flag is on AND `Current.user` is non-admin
- Patched: `ContactSerializer`, `ConversationSerializer`, `MessageSerializer`, `NotificationSerializer`, `Contact#push_event_data`, `Message#conversation_push_event_data`, and 4 legacy jbuilders
- Guards via `Current.user.administrator?` (covers `super_admin` / `account_owner` / `administrator` / `admin`)
- Background jobs / webhooks / service tokens see clean data (Current.user is nil → masker bails out)

## Out of scope (registrar pro PM)
- `contact_inboxes.source_id` agora mascarado, mas o frontend depende dele pra wa.me/click-to-call — atendente que precisar ligar terá UX degradada (esperado: o objetivo é justamente impedir contato fora)
- Widget endpoints (`/widget/contacts/*`) — não mascarados (contato vendo o próprio perfil)
- Webhooks externos — não mascarados (integrações dependem)

## Test plan
- [ ] `bundle exec rspec spec/lib/contact_pii_masker_spec.rb` → 21/21 passing
- [ ] As non-admin agent: GET `/api/v1/contacts` → phone/email/identifier mascarados
- [ ] As non-admin agent: GET `/api/v1/conversations` → meta.sender masked, contact masked, last_non_activity_message.sender masked
- [ ] As admin: same endpoints → all clean
- [ ] Webhook delivery to external URL → still receives clean data

## Linked Issue
- EVO-1551


dpaes:
## 🔴 Reprovado — EVO-1551 (vazamento de PII no ActionCable)

Bloqueador de segurança confirmado neste PR. Card volta para **Todo**; PR fica aberto.

### CB-2 — PII crua no frame WebSocket pro agente
`ContactPiiMasker.should_mask?` retorna false quando `Current.user`/`Current.account` é nil. O `ActionCableListener#message_created` broadcasta `message.push_event_data` pros tokens de todos os agentes do inbox. Para a **mensagem recebida do contato** (o evento realtime mais comum), o processamento roda em **job de background** (`Webhooks::WhatsappEventsJob` → `incoming_message_service`, que nunca seta `Current.*`). Logo, mesmo com a flag ON, o frame WS `message.created` carrega `phone/email/identifier/name` **crus** — exatamente a "Network-tab leak" que o masker afirma fechar. A carga HTTP inicial vem mascarada; o push ao vivo vaza.

**Trilha:** `app/lib/contact_pii_masker.rb` (`return false if user.nil?`); `app/listeners/action_cable_listener.rb`; `app/jobs/webhooks/whatsapp_events_job.rb` (sem `Current.`); `app/models/message.rb:353` + `~157-165`; `app/dispatchers/sync_dispatcher.rb`.

**Fix sugerido:** resolver o contexto de conta no caminho de broadcast independentemente de `Current.user` (mascarar no `push_event_data` com base na config da conta, não na presença de um user logado).

**Além disso:** este PR está **BEHIND develop** (1 commit, `011ec02` não-relacionado) — precisa rebase antes do merge.

dpaes (changes requested):
## 🔴 Changes requested — EVO-1551 round 5 (re-review)

5-dimension adversarial re-review (root-cause, egress sweep, B3, spec rigor, auth/misc), every blocker/high refute-verified by an independent verifier. The round-4 work is mostly solid — but **one blocker** holds the merge, and it's the same recurring failure mode that rejected rounds 1-4: a REST fix whose WebSocket twin was missed.

### 🔴 Blocker — `content_attributes` ships RAW on the WebSocket broadcast
`Message#push_event_data` (`app/models/message.rb:142`) builds the frame via `attributes.symbolize_keys.merge(...)`. `attributes` includes the `content_attributes` store column (`message.rb:104`: `submitted_email`, `submitted_values`, `email`), so the WebWidget pre-chat captured email/phone ship **raw** in every `message.created` / `message.updated` / `first_reply_created` frame, broadcast to all `user_tokens` (`action_cable_listener.rb:38,182-185`) — i.e. non-admin agents on the inbox. The H2 scrub (`scrub_pii_content_attributes`) is wired ONLY into the two REST serializers (`message_serializer.rb:37`, `conversation_serializer.rb:209`), never into `push_event_data`. Frontend reads `data.content_attributes` verbatim off the frame (`WebSocketContext.tsx:271`) and UI masking is visual-only.

**Repro:** flag ON, a lead submits the WebWidget pre-chat form → a non-admin agent opens DevTools → Network/WS and recovers the lead's raw email + phone from the live frame. Exactly the leak this card exists to close. No spec exercises `content_attributes` on any `push_event_data` path, so a shallow smoke won't catch it.

This is independent of the round-4 root-cause fix (which is correct) — the masker is simply never invoked for this field on the WS path.

**Fix path:** route `content_attributes` through `ContactPiiMasker.scrub_pii_content_attributes` in `push_event_data` (and the conversation embed via `EventDataPresenter#push_messages`) when `account_flag_enabled?` — the same predicate you already use for `source_id` at `message.rb:162` (CB-6) — plus a spec asserting the scrub on the WS frame.

### ✅ Solid (keep — round 4 delivered the rest)
- **Root cause B1/B2** genuinely closed: `resolved_account → RuntimeConfig.account` closes the "Current.account nil on a worker thread" class at the predicate; async is ActiveJob-only (no raw threads/Kafka/RabbitMQ); `around_perform` is reentrancy-safe and never clobbers an outer value. The worker-shape spec (`action_cable_broadcast_job_spec.rb:54-104`) is real — uses `Current.reset`, leaves `Current.account` nil, stubs only `RuntimeConfig`, asserts masked output. Proves the fix.
- **B3** contactable_inboxes: allowlist exposes only opaque non-PII ids, default-deny strips phone-bearing channels, the conversation-creation round-trip doesn't break (builder regenerates `source_id` server-side, byte-identical), admin sees raw. Couldn't break it.
- **Auth gate (CB-1)**: effective-change + deep_merge sound, no partial-PATCH bypass.
- **R5-L3 → EVO-1693**: deferral is legit — the role-key footgun only over-masks an admin (UX), never makes an agent see raw and never fails open. Contract intact in the leak direction.
- **Model integrity**: the masker never mutates the model → WhatsApp/automation/AI/webhooks read raw, delivery unaffected.

### 🟡 Non-blocking (not holding beyond the blocker)
- **Medium**: `mask_phone_like_name` leaves an embedded phone raw on mixed names (any letter short-circuits to raw, e.g. `"Cliente 11999998888"`). Bounded (dedicated phone/identifier fields are masked), but a residual leak in `name`.
- **Low**: no regression spec for `content_attributes` on the WS path (needed once fixed); the CB-2/3/4 listener specs still stub `Current.account` (SR-1 is what actually proves the fallback); `around_perform` has no own spec; specs not run here (Docker down) + crm has no rspec in CI → please run `bundle exec rspec` locally and paste the output.
- **Low/note**: `additional_attributes` ships raw in `push_event_data`/`ContactSerializer` (outside the card's field set, pre-existing).

Card → Todo. PR stays open. Pinged you on Slack with the simplified version. 🙏

marcelogorutuba:
Round 6 pushed (25d1680).

## Daniel's round 5 blocker — closed
Raw `content_attributes` on the WebSocket frame: `Message#push_event_data` now goes through `content_attributes_for_egress(audience: :broadcast)`, which uses `account_flag_enabled?` — same predicate as the `source_id` mask at `message.rb:191`. Frame is asserted clean by the new regression spec.

## Beyond the blocker — full egress closure
Routed every other shape that could carry `content_attributes` past a server boundary through the same masker entrypoint:
- `Message#webhook_data` (outbound webhook payload)
- `MessageSerializer`, `ConversationSerializer` (consolidated, no more inline `if should_mask?`)
- `_message.json.jbuilder`, `_widget_message.json.jbuilder`, `widget/messages/{create,index}.json.jbuilder`, `conversations/_conversation.json.jbuilder`

## Regression armed at CI
- **Rubocop cop** `NoRawContentAttributesInEgress` bans raw `.content_attributes` and `attributes[:content_attributes]` (the round-5 bug shape) inside serializers / jbuilder views / `message.rb` / webhook+actioncable listeners. Future paths fail CI.
- **Invariant spec** `spec/regression/pii_egress_invariant_spec.rb` walks every registered egress shape against a message carrying every PII-bearing key and asserts no PII token reaches the wire. A second meta-test greps cop-scoped paths to fail if any raw `.content_attributes` slips through.

## Round 5 nit — closed
`mask_phone_like_name` now masks the embedded phone in mixed strings like `"Cliente 11999998888"` → `"Cliente *******8888"`, instead of returning raw on any letter.

## Bonus audit finding — closed
`evolution_go_handlers/messages_upsert.rb:254` logged the full `content_attributes` hash to Rails logger — would have leaked `submitted_email`/`submitted_values` to log aggregators. Now logs keys only.

## Not addressed
The "uns nits de spec" you mentioned in round 5 weren't enumerated in the summary message. If you can drop them inline on this round 6 push, I'll fold them in.

dpaes (approved):
## Code review — EVO-1551 lead data masking — ✅ Approved (round 6)

Round 6 closes the single round-5 blocker (raw `content_attributes` in the WebSocket broadcast) — verified at the root, not patched at one call site.

- `Message#push_event_data` now overrides the raw column with `content_attributes_for_egress(audience: :broadcast)` → `account_flag_enabled?` (an admin caller can't defeat masking on a shared broadcast). The conversation embed (`EventDataPresenter#push_messages` → `message.push_event_data`) **inherits the scrub automatically** — there is no separate WS twin left to miss.
- REST serializers + the 5 jbuilders route through `content_attributes_for_egress(audience: :per_request)` (admin raw / agent masked).
- The whack-a-mole is structurally closed: a single chokepoint, a Rubocop cop (`NoRawContentAttributesInEgress`) that bans `.content_attributes` sends and the round-5 `attributes[:content_attributes]` bug shape in egress files, and an invariant spec doing both a runtime walk (PII tokens never appear in any egress output) and a static raw-read ban.

**Non-blocking notes:**
1. `webhook_data` now also scrubs `content_attributes` (registered in EGRESS_PATHS; outbound webhook classified as broadcast audience). This tightens beyond the earlier "webhooks out of scope" stance. Only the PII sub-keys are scrubbed and only when the opt-in flag is on, so it is defensible for the threat model — but confirm it won't break an integration that consumes the pre-chat `submitted_email`/`submitted_values`.
2. The cop + invariant spec are local guards: crm-community CI runs Sourcery/staleness/contract only (no RSpec/Rubocop), so the regression net is not CI-enforced. Wiring it into CI would be a good follow-up. The 0-offenses run is self-reported.

frontend#147 + auth#39 are unchanged since round 4 (round 6 is crm-only), previously vetted solid, now MERGEABLE/CLEAN. Approving and squash-merging the 3 PRs.

PR #138, opened 2026-06-11
feat(templates): wire message-send call sites to global MessageTemplate (EVO-1235)

daniloleonecarneiro:
## Summary

Story 6.6 — wire every message-send call site to the global `MessageTemplate` model introduced in 6.2/6.4. Until now those templates (`channel_id IS NULL`) were **unreachable by any send path** (`MessageBuilder#process_template_content` only looked at the inbox's channel-bound templates) and resolution was name-only. This makes `message_template_id` the canonical, global-aware resolution key across the send paths, keeping the legacy name/inline flows working during a coexistence period.

> Re-scope note (4th story in the epic with an outdated premise): the ticket's campaign composer + inbox reply are already template-based (campaign send lives in evo-flow; this CRM has no campaigns table), and the "public API" it implies is inbound-only. The real, in-repo gaps were: global templates unreachable, name-only resolution, and no public *outbound* templated send. This PR closes those.

## Changes

- **`MessageTemplates::SendResolver`** (new): id-first (global or same-channel), fallback name+language, global-aware.
- **`MessageBuilder`**: resolves by `message_template_id` (fallback name), id-only payloads allowed, persists the resolved id in `additional_attributes.template_params` (with `name`, required by `Message::TEMPLATE_PARAMS_SCHEMA`).
- **New public outbound send action** on the `Channel::Api` surface (`PublicController` + identifier auth): renders `message_template_id`; `422` on unknown id / missing required variable.
- **Legacy WARN** on inline `content` (consumer id via `X-Client-ID` → request IP) on both public endpoints — coexistence, not removal.
- **Greeting / Out-of-Office** auto-replies can reference a template (new nullable inbox columns `*_message_template_id` + relaxed hook guards), falling back to the inline string.
- **Agent-bot reply** + **macros/automation** send actions accept a template id (automation maps legacy `template_id` → `id`).

## Migration

`20260611130000_add_template_refs_to_inboxes` — adds nullable uuid `greeting_message_template_id` + `out_of_office_message_template_id`. Reversible. (Version after develop's latest.)

## Test plan

- New/extended specs (22 examples, green on the test DB): `send_resolver_spec`, `message_builder_spec`, public `outbound_messages_controller_spec` (id→render, inline→WARN, unknown id→422, missing required var→422), inbound WARN spec, `hook_execution_service` greeting/OOO, `macros/execution_service` send_message.
- `rubocop` clean on changed files (no net-new offenses vs develop baseline).
- Build templates inline (no factory); required-var tests set `variables:` explicitly.

## Notes

- Pairs with the frontend PR (evolution-foundation/evo-ai-frontend-community) that sends `id` in `template_params` so the live inbox path references `message_template_id`.
- Legacy inline API kept (deprecated + WARN) for the transition.

### DIFF
--- a/app/builders/messages/message_builder.rb
+++ b/app/builders/messages/message_builder.rb
@@ -158,22 +158,28 @@ class Messages::MessageBuilder
     return unless @params[:template_params].present?
 
     template_info = @params[:template_params].is_a?(Hash) ? @params[:template_params] : JSON.parse(@params[:template_params].to_json)
-    return unless template_info['name'].present?
+    template_id = template_info['id'] || @params[:message_template_id]
+    # Allow id-only payloads: the canonical key is the id, name is the fallback.
+    return unless template_info['name'].present? || template_id.present?
 
-    Rails.logger.info "Processing template: name=#{template_info['name']}, language=#{template_info['language']}, inbox_type=#{@conversation.inbox&.inbox_type}"
+    Rails.logger.info "Processing template: id=#{template_id}, name=#{template_info['name']}, language=#{template_info['language']}, inbox_type=#{@conversation.inbox&.inbox_type}"
 
-    # Find template by name and language
-    template = @conversation.inbox.channel&.message_templates&.active&.find_by(
+    # Resolve id-first (global-aware), falling back to name + language.
+    template = MessageTemplates::SendResolver.new(
+      id: template_id,
       name: template_info['name'],
-      language: template_info['language'] || 'pt_BR'
-    )
+      language: template_info['language'],
+      channel: @conversation.inbox&.channel
+    ).resolve
 
     unless template
-      Rails.logger.error "Template not found: name=#{template_info['name']}, language=#{template_info['language'] || 'pt_BR'}, channel_id=#{@conversation.inbox.channel&.id}"
+      Rails.logger.error "Template not found: id=#{template_id}, name=#{template_info['name']}, language=#{template_info['language'] || 'pt_BR'}, channel_id=#{@conversation.inbox.channel&.id}"
       return
     end
 
     Rails.logger.info "Template found: #{template.name} (#{template.id}), metadata keys: #{template.metadata&.keys&.inspect}"
+
+    persist_resolved_template_id(template)
 
     # Process template based on channel type
     if @conversation.inbox&.inbox_type == 'Email'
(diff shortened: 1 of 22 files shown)


dpaes (approved):
## Code review — EVO-1235 [6.6] · ✅ Approved (with non-blocking notes)

Reviewed against the Linear acceptance criteria, with the re-scope verified against the actual code.

### Re-scope verified
- **No campaign composer in this CRM** — confirmed: no `campaigns` table in `schema.rb`, no campaign route/controller; `Campaign.rb` is only a lookup shim for the EvoFlow enrich path. So AC1 (campaign dropdown) is genuinely out-of-repo (lives in evo-flow). Accepted.
- **Global templates were unreachable** — confirmed: the old `MessageBuilder#process_template_content` only looked at `inbox.channel.message_templates.find_by(name:)`, so `channel_id IS NULL` templates and id-based lookup were excluded. This PR closes that real gap.

### AC mapping
- AC1 (campaign composer) → out-of-repo (evo-flow); not in scope here.
- AC2 (inbox "insert template") → pre-existing modal + FE now sends `id`. ✅
- AC3 (public API `message_template_id` → templated send) → ✅ new `OutboundMessagesController`, spec'd.
- AC4 (public API inline → send + WARN) → ✅ on the outbound endpoint, spec'd.
- AC5 (specs green, no regression) → 22 specs reported green (note: this repo does not run RSpec in CI — please keep confirming with a local `bundle exec rspec` run).

### What's solid
- `SendResolver` is clean: id-first (global or same-channel), name+language fallback with channel→global priority, nil on miss; good spec coverage incl. cross-channel visibility, inactive, missing id.
- Schema-safe: `Message::TEMPLATE_PARAMS_SCHEMA` requires only `name` and has no `additionalProperties: false`, so the new `id` key validates and `name` is backfilled on every path (`persist_resolved_template_id`, outbound, etc.). `MessageBuilder` order is correct — `@message` built before `process_template_content`, content from `render_with_variables`, `name` set before `save!`.
- Invalid (non-UUID) `message_template_id` on the outbound endpoint casts to `IS NULL` → nil → clean `422` (no 500).

### Non-blocking findings (follow-up, not blocking merge)
1. **MEDIUM — inbound `messages#create` WARN is a false positive.** Inbound (contact→agent) messages always carry `content` and have no template alternative, so `warn_legacy_inline_content` fires on every normal inbound message. That's log spam and it pollutes the very signal meant to identify legacy *outbound* consumers. Suggest removing the WARN from the inbound controller (keep it on `outbound_messages#create` only).
2. **LOW — dormant capabilities.** `agent_bots/message_creator#create_bot_reply`, macro `send_message`, and automation `send_template` now *accept* a template id, but no caller/producer passes it yet (`response_processor`, `bot_runtime_controller`, `n8n_request_service` and the macro/automation builders are unchanged). Capability is ready but unreached until a producer supplies the id — flagging so it isn't mistaken for a live path.
3. **LOW — auto-reply blank edge.** If a greeting/OOO template has an explicitly `required: true` variable outside `{first_name, name, email}`, render raises → fallback to the inline string; if that column is blank, an empty greeting can go out. Narrow, but worth a guard against blank rendered content.

Approving on the merits — the in-repo gaps are closed, well-tested and well-documented. The re-scope (AC1 out-of-repo) was accepted by the reviewer.

PR #41, opened 2026-05-05
fix(crm): credential fallback + eager webhook for Evolution Go (EVO-984)

nickoliveira23:
## Summary

Fixes the two bugs reported in [EVO-984](https://linear.app/evoai/issue/EVO-984):

1. **Legacy Evolution Go channels with empty `provider_config['api_url']`/`['admin_token']` fail every API call.** The 4 controllers that operate on existing channels (`qrcodes`, `settings`, `privacy`, `profile`) read those fields directly with no fallback. Operators see "Erro ao gerar QR Code" and have to recreate the channel from scratch.
2. **On instance creation, the webhook URL is never sent to Evolution Go.** `connect_instance` only fires when the user opens the QR screen — any inbound message between create and QR is dropped.

## Changes

- New `Whatsapp::EvolutionGoConcern#evolution_go_credentials_for(channel)` helper resolves `api_url`/`admin_token` from `provider_config`, falling back to `GlobalConfigService.load('EVOLUTION_GO_API_URL' / '_ADMIN_SECRET')` when blank. Logs `Rails.logger.warn` when the channel is found but `instance_token` is blank, so corrupt-channel cases are observable in production.
- The 4 controllers (`qrcodes`, `settings`, `privacy`, `profile`) now use the helper in `set_instance_params` instead of reading `provider_config` directly. `privacy`/`profile`/`settings` additionally `include EvolutionGoConcern` (`qrcodes` already had it).
- `AuthorizationsController#create` now invokes `register_webhook_after_create(api_url, instance_token)` right after `create_instance_go` succeeds, so the webhook is registered eagerly. Failures are logged and swallowed to avoid regressing channel creation — the QR action also re-runs the same `connect_instance` call later.

## Validation

- `evo-ai-crm-community: bundle exec rspec spec/controllers/concerns/evolution_go_concern_spec.rb spec/controllers/api/v1/evolution_go/` — **16 examples, 0 failures**
- `evo-ai-crm-community: ruby -c` on every changed file — clean
- Manual end-to-end (Rails console + browser):
  - Legacy channel with `update_column(:provider_config, ...empty...)` → Connect generates QR (was "Erro ao gerar QR Code" pré-fix). Fallback resolved via `EVOLUTION_GO_API_URL`/`EVOLUTION_GO_ADMIN_SECRET` from Admin Settings.
  - Fresh channel creation → Evolution Go logs show `Creating instance` + `Connecting instance at .../instance/connect` in the same request, before user opens QR.

### Backend
[...]

PR #140, opened 2026-06-12
feat(message-templates): dedicated CRUD endpoint + inbox cutover (EVO-1716)

daniloleonecarneiro:
## Summary

EVO-1716 closes the loop opened at the start of the Templates epic: a dedicated, account-scoped endpoint for the template save flow, instead of the inbox-nested member actions + `?global=true` toggle bolted onto `InboxesController`.

- **New `Api::V1::MessageTemplatesController`** at flat `/api/v1/message_templates`, serving both:
  - **global** (channel-less) templates — the default when no inbox is given;
  - **channel-bound** templates — when `inbox_id` (or `channel_id`) is present, the op delegates to the channel, preserving the WhatsApp Cloud writeback (`create_template` → Meta) **exactly as before**.
- **New `MessageTemplatePolicy`** + **`MessageTemplateParams`** concern (the permitted shape was extracted verbatim from `InboxesController`, now single-sourced).
- **Routes**: flat `resources :message_templates`; removed the 4 inbox CRUD member routes. **Meta sync stays inbox-scoped** (`sync` / `sync_with_whatsapp_cloud` kept — per-channel by nature).
- **`InboxesController`**: −360 lines (CRUD methods + orphaned helpers + 3 stale `require_permissions` entries).

Templates are an **instance-wide catalog** (no `account_id`; this CRM runs against a single runtime account), so there is no per-record account scoping to add.

## Test plan

- `spec/requests/api/v1/message_templates_spec.rb` — global + channel-bound CRUD, global-vs-bound isolation, duplicate-name 422, authorization 403, legacy-route cutover 404. **22 examples, all green.**
- `spec/requests/api/v1/message_templates_service_token_spec.rb` — evo-flow service-token journey via `inbox_id` filter.
- Inbox controller sync regression specs: 3/3 green. Rubocop clean on touched production files.

## Notas

- **Deploy ordering**: the auth-service RBAC PR (`message_templates` resource) **must ship before** this backend, otherwise non-admin users lose template CRUD access.
- Companion PRs: auth-service (RBAC resource) and frontend (repoint to flat endpoint).

## Summary by Sourcery

Introduce a dedicated, account-scoped message templates API and cut over from inbox-nested CRUD while preserving per-channel Meta sync.

New Features:
- Add Api::V1::MessageTemplatesController providing global and channel-bound CRUD via a flat /api/v1/message_templates endpoint.
- Introduce MessageTemplatePolicy and MessageTemplateParams concern to centralize authorization and strong parameters for templates.

Enhancements:
- Remove message template CRUD logic from InboxesController and simplify inbox routes to only keep per-channel Meta sync actions.
- Update service-token and request specs to cover the new message templates endpoint, including authorization, scoping, and legacy route cutover behavior.

Tests:
- Expand and adjust request specs for message templates to cover global and channel-bound CRUD, isolation rules, authorization, and WhatsApp Cloud sync regression.
- Update service token request specs to validate access to channel templates through the new endpoint.

### DIFF
--- a/app/controllers/api/v1/inboxes_controller.rb
+++ b/app/controllers/api/v1/inboxes_controller.rb
@@ -28,11 +28,10 @@ module Api
           disconnect_channel_provider: 'inboxes.update',
           sync_whatsapp_subscription: 'inboxes.update',
           avatar: 'inboxes.update',
-          message_templates: 'inboxes.message_templates',
+          # Template CRUD moved to MessageTemplatesController (EVO-1716); only the
+          # per-channel Meta sync remains here, keeping its inbox permission.
           sync_message_templates: 'inboxes.message_templates',
           sync_template_with_whatsapp_cloud: 'inboxes.message_templates',
-          update_message_template: 'inboxes.update_message_template',
-          delete_message_template: 'inboxes.delete_message_template',
           facebook_posts: 'inboxes.read'
         })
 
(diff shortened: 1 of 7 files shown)


dpaes (approved):
## ✅ Approved — EVO-1716 (CRM: dedicated endpoint + inbox cutover)

Reviewed against source across all three repos, not just the diff.

**Verified**
- **Clean cutover:** no dangling references to the removed helpers/actions remain in `InboxesController` (head, 818 lines); the kept methods (`sync_message_templates`, `sync_template_with_whatsapp_cloud`, `extract_template_params`, `render_template_name_required_error`) are intact.
- **No cross-channel write:** `channel.update_message_template` / `delete_message_template` resolve via `message_templates.find(id)` on the scoped association → a mismatched id 404s; the global path scopes to `channel_id: nil`, so it can't reach channel-bound rows on update/destroy.
- **WhatsApp Cloud writeback preserved** verbatim in `build_template` (`create_template` → Meta); Meta sync correctly stays inbox-scoped.
- **Authorization:** the real 403 gate is `require_permissions` → `EvoAuthService#check_user_permission('message_templates.*')`. (The Pundit `MessageTemplatePolicy` is effectively open in the community edition, since `User#has_permission?` always returns `true` — defense-in-depth only.) The matching auth-service resource + grants ship in #41.
- No schema migration; the change is internal wiring only.

**Non-blocking (follow-up at your discretion — not holding merge):**
1. _(Medium)_ Happy-path authz for a real non-admin user is untested — every green CRUD spec uses a service token, which bypasses both gates. A single "agent **with** `message_templates.create` → 201" spec (the remote check is mocked anyway) would lock the new RBAC contract.
2. _(Low)_ The `sync_with_whatsapp_cloud` block dropped 2 still-applicable examples ("422 for a non-WhatsApp-Cloud channel", "404 for an unknown id") on an endpoint this PR preserves (5 → 3).
3. _(Low)_ `show` / global `index` aren't channel-scoped — any read-permitted user can fetch any template by id. Consistent with the instance-wide-catalog decision; no FE caller uses `show`.
4. _(Low)_ `resolve_channel` via `channel_id` alone uses `find_by!` → a `channel_id`-only create/update/destroy with no existing row 404s as "Inbox not found". Latent (no client hits this path; FE always uses `inbox_id`).
5. _(Nit)_ `MessageTemplatePolicy::Scope` is dead code — `policy_scope` is never called.

Tidy refactor. LGTM.

PR #141, opened 2026-06-12
test(message-templates): review fixes left out of #140 squash (EVO-1716)

daniloleonecarneiro:
## Summary

Follow-up a #140. As correções da adversarial review foram commitadas **depois** que #140 já tinha sido mergeado (squash levou só o primeiro commit, head `f7b4440`), então ficaram fora do develop. Este PR traz exatamente esse delta.

- **Authz happy-path com user real**: todo verde anterior rodava via service-token, que bypassa os dois gates (`require_permissions` + Pundit). Autentica um user e stuba `check_user_permission` **por key** para travar o contrato do resource novo (`create` → 201, `index` → 200, negado → 403).
- **Cobertura de sync restaurada** no `sync_with_whatsapp_cloud` (preservado, não movido): **422** para canal não-WhatsApp-Cloud e **404** para template id inexistente.
- **`MessageTemplatePolicy`**: removida a classe `Scope` (dead code — `policy_scope` nunca é chamado sobre `MessageTemplate`; o controller filtra via `base_scope`).
- **404 enganoso**: quando o caller manda `channel_id` puro, a mensagem agora nomeia o alvo certo ("Channel not found") em vez de "Inbox not found".

## Test plan

- `message_templates_spec.rb` + `message_templates_service_token_spec.rb`: **26 examples, 0 failures**.
- Rubocop limpo nos arquivos de produção (controller + policy).

## Summary by Sourcery

Align message template authorization, sync behavior, and error responses with the intended contract and restore missing test coverage from a previous squash merge.

Bug Fixes:
- Return 404 errors that correctly distinguish between missing inboxes and channels when resolving message template targets.
- Ensure sync_with_whatsapp_cloud returns 422 for non-WhatsApp-Cloud channels and 404 for unknown template IDs instead of silently proceeding.

Enhancements:
- Remove the unused Pundit Scope from MessageTemplatePolicy since templates are not account-scoped and the controller does not call policy_scope.
- Expand authorization tests to exercise the real-user path using EvoAuthService permission checks for message_templates.create and message_templates.read.

Tests:
- Add request specs covering real-user authorization for message template create and index endpoints, including allowed and forbidden scenarios.
- Add request specs verifying sync_with_whatsapp_cloud behavior for non-WhatsApp-Cloud channels and unknown template IDs.

### DIFF
--- a/app/controllers/api/v1/message_templates_controller.rb
+++ b/app/controllers/api/v1/message_templates_controller.rb
@@ -220,8 +220,12 @@ module Api
                        status: :unprocessable_entity)
       end
 
+      # The channel target failed to resolve. Name the param the caller actually
+      # sent (inbox_id resolves an inbox; a bare channel_id resolves the channel
+      # off an existing row) so the 404 message isn't misleading.
       def render_inbox_not_found
-        error_response(ApiErrorCodes::RESOURCE_NOT_FOUND, 'Inbox not found', status: :not_found)
+        target = params[:inbox_id].present? ? 'Inbox' : 'Channel'
+        error_response(ApiErrorCodes::RESOURCE_NOT_FOUND, "#{target} not found", status: :not_found)
       end
     end
   end
--- a/app/controllers/api/v1/message_templates_controller.rb
+++ b/app/controllers/api/v1/message_templates_controller.rb
@@ -30,7 +30,7 @@ module Api
                           })
 
       # GET /api/v1/message_templates
-      # Filters: inbox_id (resolve inbox→channel), channel_id, or neither (global).
+      # Filters: inbox_id (resolve inbox→channel) or neither (global catalog).
       # Plus category / template_type / search / sort_by / page / per_page.
       def index
         authorize MessageTemplate, :index?, policy_class: MessageTemplatePolicy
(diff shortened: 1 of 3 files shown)


dpaes (approved):
**Reviewed — approved ✅**

Verified the follow-up delta against the real code on `danilocarneiro/evo-1716-review-fixes`. CRM-community runs no RSpec in CI, so I traced each spec's logic by hand rather than trusting the self-reported green.

- **Real-user authz happy-path** — traced the full chain: `require_permissions` → `check_<action>_permission!` → `check_permission!(key)` → `EvoAuthService#check_user_permission(user_id, key)` (2-arg signature, matches `.with(anything, 'message_templates.create')`). Exactly one call per request with the action's key; Pundit goes through `User#has_permission?` (always true on community), not EvoAuthService, so the strict `.with` stub can't trip "unexpected arguments". 201 / 200 / 403 are all sound.
- **Sync coverage restored** — `sync_template_with_whatsapp_cloud` does a global `MessageTemplate.find` → unknown uuid raises `RecordNotFound` → 404; the `Channel::Whatsapp && provider == 'whatsapp_cloud'` guard rejects the evolution channel → 422 (the cross-channel template is still found globally, hence 422 and not 404 — the fixtures line up).
- **`MessageTemplatePolicy::Scope` removal** — confirmed zero callers of `policy_scope(MessageTemplate)` anywhere; the only `policy_scope` reference is the generic `Pundit.policy_scope!` in `application_policy.rb`. The controller authorizes instances + filters via `base_scope`, so the Scope class was dead code. Safe to drop.
- **`channel_id` surface removal** — gone from the controller entirely; `base_scope` / `channel_bound?` / `resolve_channel` collapse to the single `inbox_id` axis. Confirmed no consumer ever sent `channel_id` (FE global service sends neither filter, channel-bound sends `inbox_id`; evo-flow reads the table directly), so the "Inbox not found" 404 is now always accurate.

Non-blocking nit: the PR description's "404 → 'Channel not found'" bullet describes commit 1's intermediate state — commit 2 dropped `channel_id` entirely, so the net diff only keeps the always-correct "Inbox not found". The code itself is coherent.

The one open item from the card (`show` / `index` global without channel scope) is an accepted by-design consequence of the account-less, instance-wide catalog — not a blocker.

PR #148, opened 2026-06-15
fix(templates): preserve same-named legacy migrations + harden run (EVO-1718)

daniloleonecarneiro:
## Summary

Non-blocking follow-up to the review of the EVO-1234 [6.5] legacy → `MessageTemplate` migration job (`MigrateLegacyTemplatesToMessageTemplateJob`). The feature was already correct (dry-run + rollback + idempotent); this addresses the review findings:

- **🟡 [Medium] Duplicate name → data loss.** When two channel-coupled rows shared a name (different channels), only the first became a global; the second was skipped `:duplicate_name` **forever**, so its content never reached the global menu. `resolve_name` now **preserves both**: a legacy-vs-legacy collision is suffixed (`"<name> (legacy)"`, then `"<name> (legacy <id>)"`), never skipped. The dead `:duplicate` sentinel, `REASON_DUPLICATE_NAME`, the `legacy_names` dedup set and `legacy_global_exists?` are removed (the latter is subsumed by `global_name_taken?`). The two-tier admin-collision suffix is unchanged.
- **🟡 [Medium] `media_type` had no enum guard.** Added `normalized_media_type` mirroring `normalized_template_type`, wired into `build_attrs` (replacing `source.media_type.presence`). **Note:** in Rails 7.1 the enum reader already deserializes an out-of-enum stored value to `nil` and assignment does not raise — so this is **defensive symmetry, not a behavior fix** (the prior "raises ArgumentError on assignment" comment was wrong and is corrected). The rescue now logs `record.errors.full_messages.presence || e.message`, never an empty string.
- **🟡 [Medium] Rake exited 0 even when every row failed.** `templates:migrate_legacy` now `abort`s (non-zero exit) when `summary[:skipped][:error] > 0`, **after** printing the summary. Gates on `:error` (unexpected exceptions) only, not `:invalid` (a predictable dry-run validation skip); applies to dry-run too. `rollback_legacy_migration` is unchanged.

No DB migration, no destructive change — originals and admin globals are never touched.

## Test plan

`spec/jobs/migrate_legacy_templates_to_message_template_job_spec.rb` — **13 examples, 0 failures** (run locally; crm-community does not run RSpec in CI):

- The two same-name cases reworked to **preserve-both** (dry-run predicts 2 migrations; a real run creates `Dup` + `Dup (legacy)`, nothing skipped).
- `normalized_media_type` unit-tested directly (`'sticker'`/`''`/`nil` → nil, `'image'` passes) — a DB round-trip would be a tautology given the Rails 7.1 reader.
- `source_label` tested via `instance_double(MessageTemplate, channel_type:)` (a mismatched polymorphic fixture would pass for the wrong reason).
- Rescue diagnostics: a `RecordInvalid` is logged with `full_messages` and bucketed under `:error`.
- Synthetic-variable pruning (`var_1` injected via `update_columns` to mimic the raw Dec-2025 SQL write) is dropped on the copy.
- Existing EVO-1234 ACs (idempotency, rollback scope, WhatsApp Cloud, blank content) remain green.

Rake gate verified out-of-spec (rake tasks aren't loaded there): forced `skipped[:error]=2` → prints summary then `FAILED: …` and **exit 1**; clean run and dry-run → **exit 0**. RuboCop: no new offenses on touched files.

## Notas

- **Behavior change:** `skipped[:duplicate_name]` disappears for legacy-vs-legacy collisions (now suffixed). Any dashboard keying on that bucket will see it drop to 0 — expected.
- **Bare-name "winner" is arbitrary** (UUID/PK order) but now cosmetic only — both contents survive. Left non-deterministic by design.
- The `channel_id` error-key rename mentioned in the EVO-1717 review is **EVO-1717 #2**, not this ticket; its client-facing benefit unblocks via 6.4 / EVO-1233 (already merged).

### DIFF
--- a/app/jobs/migrate_legacy_templates_to_message_template_job.rb
+++ b/app/jobs/migrate_legacy_templates_to_message_template_job.rb
@@ -35,13 +35,14 @@ class MigrateLegacyTemplatesToMessageTemplateJob < ApplicationJob
   }.freeze
   DEFAULT_SOURCE_LABEL = 'other_legacy_template'
 
-  # Skip reasons. The four below are the documented taxonomy; specs/ACs assert on
-  # these exact keys. `:invalid` (built copy failed validation, e.g. a stray
+  # Skip reasons. The three below are the documented taxonomy; specs/ACs assert
+  # on these exact keys. `:invalid` (built copy failed validation, e.g. a stray
   # media_type) and `:error` (unexpected exception) are diagnostic buckets.
+  # NOTE: there is deliberately no :duplicate_name reason — same-named legacy
+  # rows are preserved (the later one is name-suffixed), never skipped (EVO-1718).
   REASON_WHATSAPP_CLOUD = :whatsapp_cloud
   REASON_INVALID_CONTENT = :invalid_content
   REASON_ALREADY_MIGRATED = :already_migrated
-  REASON_DUPLICATE_NAME = :duplicate_name
 
   def perform(dry_run: false)
     summary = { migrated: Hash.new(0), skipped: Hash.new(0), dry_run: dry_run }
@@ -51,8 +52,7 @@ class MigrateLegacyTemplatesToMessageTemplateJob < ApplicationJob
       dry_run: dry_run,
       summary: summary,
       seen_keys: Set.new,    # external_legacy_id keys produced this run
-      all_names: Set.new,    # every global name produced this run (downcased)
-      legacy_names: Set.new  # global names produced by THIS migration this run (downcased)
+      all_names: Set.new     # every global name produced this run (downcased)
     }
 
     MessageTemplate.where.not(channel_id: nil).find_in_batches(batch_size: BATCH_SIZE) do |batch|
@@ -60,7 +60,7 @@ class MigrateLegacyTemplatesToMessageTemplateJob < ApplicationJob
         migrate_one(source, ctx)
       rescue StandardError => e
         Rails.logger.error(
-          "[migrate_legacy_templates] source_id=#{source.id} error=#{e.class}: #{e.message}"
+          "[migrate_legacy_templates] source_id=#{source.id} error=#{e.class}: #{error_detail(e)}"
         )
         summary[:skipped][:error] += 1
       end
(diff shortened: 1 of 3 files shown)


dpaes (approved):
## Review — EVO-1718 follow-up (non-blocking) ✅ Approved

Cross-checked all four EVO-1234 review findings against the full job, rake task, spec and the `MessageTemplate` model. All three 🟡 findings are correctly resolved and the 🔵 testing finding is substantially covered.

**F1 — preserve-both (same-named legacy rows):** Correct. `resolve_name` now escalates `bare → "(legacy)" → "(legacy <id>)"` instead of skipping the 2nd row as `:duplicate_name`. Verified for 2 and 3+ collisions, dry-run vs real-run name parity (in-run `all_names`), and cross-run re-runs. Idempotency is unaffected — it lives in the step-3 `external_legacy_id` guard, which runs *before* `resolve_name`. The partial unique index `idx_message_templates_global_name (name) WHERE channel_id IS NULL` backs the suffixing: any missed collision would fail loudly (`:error` → rake abort), never corrupt data.

**F2 — `media_type` guard / F10 — rescue logging:** Code is correct and safe. `normalized_media_type` maps any non-enum value to `nil` before assignment; `create!` only ever receives `nil` or a valid key. `error_detail` falls back to `e.message` for non-`RecordInvalid` errors and for records with no messages (never an empty string). One nit (non-blocking): the comment claiming the original *"assignment raises ArgumentError"* premise was wrong is itself inaccurate — in Rails 7.1.5.1 a genuinely out-of-enum **assignment** *does* raise (enums declared without `validate: true`). It never bites at runtime because the source reader already nils bad stored values and the guard nils anything else before assignment — so the guard is arguably *more* necessary than the comment states. Worth correcting the comment in a later pass.

**F3 — rake gate:** Correct. `templates:migrate_legacy` aborts with a non-zero exit when `skipped[:error] > 0`, after printing the summary, dry-run included. Fires only on `:error` (unexpected exceptions), not on dry-run `:invalid` predictions.

**Tests:** The two same-name specs were rewritten to preserve-both with order-independent matchers (`contain_exactly`, `values.sum`) — robust to the UUID PK. New unit tests for `normalized_media_type`, `source_label` (non-WhatsApp + default), the rescue diagnostics, and the F5 synthetic-variable pruning all exercise the real code paths (not tautologies; `create!` stub is narrowly scoped). No lingering references to the removed `:duplicate_name` / `legacy_names` / `legacy_global_exists?` anywhere in the repo, so the removal breaks nothing.

Caveat: crm-community runs no RSpec in CI, so the 13/13 is self-reported local — static trace confirms all 13 should pass as written.

Non-blocking observations sent to the author directly; follow-up at their discretion. Approving and merging.

PR #149, opened 2026-06-15
fix(templates): harden legacy-migration job per Sourcery review (EVO-1719)

daniloleonecarneiro:
## Summary

Non-blocking follow-up (6.10) to the **Sourcery review** of the EVO-1234 [6.5] legacy → `MessageTemplate` migration job. Four points raised; three applied, one intentionally declined.

- **🟡 [Medium] N+1 on the batch loop.** `MessageTemplate.where.not(channel_id: nil).find_in_batches` had no preload, so `keep_channel_bound?` issued a per-row `source.channel` query for every WhatsApp row. Added `.includes(:channel)`. (`channel` is a polymorphic `belongs_to`; `includes` keeps it a preload and preserves `find_in_batches`' default `ORDER BY id`.)
- **🟡 [Medium] Over-broad rollback DELETE.** `rollback_legacy_migration` deleted `where.not(external_legacy_id: nil)`. The column is exclusive to this migration today, but a future integration reusing it would be wiped. Scoped to this migration's provenance prefix: `external_legacy_id LIKE 'message_template:%'`, reusing `MigrateLegacyTemplatesToMessageTemplateJob::LEGACY_KEY_PREFIX`. The rollback spec mirrors the filter **and** now seeds a foreign-provenance row (`other_integration:1`) that must survive — without it the test was vacuous (passed against the old scope too).
- **🟢 [Low] Raw skip-reason symbols.** `:invalid` and `:error` escaped the `REASON_*` convention. Promoted to `REASON_VALIDATION_FAILED` / `REASON_ERROR` (used in the job, the rake `:error` gate, and the spec). Named **`REASON_VALIDATION_FAILED`, not `REASON_INVALID`**, to avoid a near-synonym collision with the existing `REASON_INVALID_CONTENT`. Symbol values are unchanged, and the comment's deliberate "documented taxonomy vs. diagnostic buckets" two-tier distinction is preserved.
- **⚪ [Declined] `require 'set'`.** Not added. Under Ruby 3.4 `Set` is autoloaded and `require 'set'` trips `Lint/RedundantRequireStatement` (verified: 1 offense, autocorrectable) — adding it would *introduce* a lint offense. This matches Sourcery's own "não procede / inócuo" verdict on this point.

No migration, no schema change, no behavior change.

## Test plan

- `POSTGRES_PORT=5433 ... bundle exec rspec spec/jobs/migrate_legacy_templates_to_message_template_job_spec.rb` → **13 examples, 0 failures**.
- **Non-vacuity check (rollback):** temporarily reverting the spec's delete to the old `where.not(external_legacy_id: nil)` makes the example FAIL on the foreign-provenance assertion — confirming the test now distinguishes the new scope.
- **RuboCop:** 9 offenses on the branch == 9 on the `develop` baseline → **zero new offenses**; `Lint/RedundantRequireStatement` = 0.

## Notes

- Builds on EVO-1234 (#136) and EVO-1718 (#148), both merged to `develop`.
- The `require 'set'` decision is the only one of the 4 points not coded; rationale above is the resolution for the Sourcery thread.

## Summary by Sourcery

Harden the legacy-to-MessageTemplate migration job and rollback task based on Sourcery review feedback, tightening rollback scoping, naming skip reasons, and reducing query overhead.

Bug Fixes:
- Restrict the rollback migration to delete only templates created by this migration using the legacy key prefix, ensuring admin and foreign-provenance templates are preserved.

Enhancements:
- Add eager loading of channels when batching message templates to avoid N+1 queries during migration.
- Introduce named constants for validation-failure and error skip reasons and update the job, rake task, and specs to consistently use them.
- Strengthen the rollback spec to assert that foreign-provenance templates survive and that only this migration's templates are removed.

### DIFF
--- a/app/jobs/migrate_legacy_templates_to_message_template_job.rb
+++ b/app/jobs/migrate_legacy_templates_to_message_template_job.rb
@@ -36,13 +36,17 @@ class MigrateLegacyTemplatesToMessageTemplateJob < ApplicationJob
   DEFAULT_SOURCE_LABEL = 'other_legacy_template'
 
   # Skip reasons. The three below are the documented taxonomy; specs/ACs assert
-  # on these exact keys. `:invalid` (built copy failed validation, e.g. a stray
-  # media_type) and `:error` (unexpected exception) are diagnostic buckets.
+  # on these exact keys. REASON_VALIDATION_FAILED (built copy failed validation,
+  # e.g. a stray media_type) and REASON_ERROR (unexpected exception) are the
+  # lower-tier diagnostic buckets — kept distinct from the taxonomy above.
   # NOTE: there is deliberately no :duplicate_name reason — same-named legacy
   # rows are preserved (the later one is name-suffixed), never skipped (EVO-1718).
   REASON_WHATSAPP_CLOUD = :whatsapp_cloud
   REASON_INVALID_CONTENT = :invalid_content
   REASON_ALREADY_MIGRATED = :already_migrated
+  # Diagnostic buckets (named distinctly from REASON_INVALID_CONTENT on purpose).
+  REASON_VALIDATION_FAILED = :invalid
+  REASON_ERROR = :error
 
   def perform(dry_run: false)
     summary = { migrated: Hash.new(0), skipped: Hash.new(0), dry_run: dry_run }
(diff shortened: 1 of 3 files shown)


dpaes (approved):
**Approved.** Clean, well-scoped follow-up. Verified all four Sourcery points against `origin/develop` + the PR head:

1. **N+1 → `includes(:channel)`** ✅ `keep_channel_bound?` reads `source.channel` (WhatsApp rows only); `belongs_to :channel, polymorphic: true, optional: true` → `includes` does a polymorphic preload and preserves the `find_in_batches` PK order. N+1 gone.
2. **Rollback blast-radius** ✅ Now scoped to `external_legacy_id LIKE 'message_template:%'` reusing `LEGACY_KEY_PREFIX` (matches `legacy_key`). The spec is non-vacuous: it seeds a foreign-provenance row (`other_integration:1`) and asserts it survives — the old unscoped delete would have wrongly removed it.
3. **`require 'set'`** ⚪ Correctly NOT added — `Set` is autoloaded on Ruby 3.2+ and the require would trip `Lint/RedundantRequireStatement`, matching Sourcery's own "não procede" verdict.
4. **Reason constants** ✅ `:invalid`/`:error` promoted to `REASON_VALIDATION_FAILED` / `REASON_ERROR` across job + rake gate + spec. Symbol values unchanged, so the summary buckets and `reason.to_s` metrics stay backward-compatible; no other reader of the raw bucket exists in the repo.

Non-blocking nit (cosmetic): the `_` chars in `'message_template:%'` are single-char LIKE wildcards, so it could in theory match `messageXtemplate:...`. Practically harmless (no plausible foreign prefix matches; rollback is a manual operator task) — no follow-up needed.

No schema/behavior change. Note: this repo's CI doesn't run RSpec, so the 13/13 green run is the author's local result.

PR #150, opened 2026-06-15
fix(templates): silence inbound WARN + guard blank auto-replies (EVO-1720)

daniloleonecarneiro:
## Summary

Non-blocking follow-up (6.11) to the **EVO-1235 [6.6]** wire-call-sites review. Three findings raised; two fixed, one documented.

- **🟡 [Medium] False-positive "deprecated inline content" WARN on INBOUND.** `Public::Api::V1::Inboxes::MessagesController#create` included `TemplateConsumerLogging` and fired `warn_legacy_inline_content` whenever `content` was present. But an **inbound** message (from the contact) always carries `content` and has **no template alternative** — there is nothing to migrate. The WARN fired on every inbound message, spamming the very signal meant to identify legacy **outbound** consumers. Removed the `include` + the call (and its coexistence comment). The `outbound_messages_controller` (the legitimate WARN site) and the `TemplateConsumerLogging` concern are untouched; `permitted_params`/`message_params` are preserved.

- **🟢 [Low] Blank auto-reply edge.** Greeting/Out-of-Office fire when `*_message_template_id` is present. If the referenced template has a `required: true` variable outside `{first_name, name, email}`, `render_with_variables` raises `ArgumentError`; `TemplateContent#template_content_for` rescues it and falls back to the inline column. `Message` has **no `content` presence validation**, so a blank inline column produced a blank `:template` message. `perform` now resolves the content **once** (via a `*_content` helper, so the guard adds no extra `SendResolver` lookup) and `return if content.blank?` — skip silently. Skipping the blank `:template` message keeps `HookExecutionService#first_message_from_contact?` accurate, so a later valid greeting can still fire.

- **🟢 [Low / Documented] Dormant `message_template_id` accept-paths** in agent-bot/macro/automation are intentionally untouched — no producer feeds the id yet (expected until a UI does).

No migration, no schema change.

## Test plan

- `POSTGRES_PORT=5433 ... bundle exec rspec spec/services/message_templates/template/greeting_spec.rb spec/services/message_templates/template/out_of_office_spec.rb spec/requests/public/api/v1/inboxes/messages_controller_spec.rb spec/requests/public/api/v1/inboxes/outbound_messages_controller_spec.rb` → **14 examples, 0 failures**.
- **Inbound spec inverted:** the existing example asserted the WARN *fired*; it now asserts the incoming message persists (`incoming`, with the posted content) **and** the deprecated WARN is **not** emitted (`allow → have_received` so unrelated warns don't interfere).
- **New service specs** build a global `MessageTemplate` with a surviving `required` variable (`{{order_id}}` in content + explicit `required: true`, since `extract_variables_from_content` strips vars absent from content) and **assert `render_with_variables` actually raises before relying on the guard** (no vacuous test). Cover: blank-skip, render-raise→inline fallback, and no-template→inline. Created fallback message is `message_type: :template`, `content_type: 'text'`.
- **RuboCop:** clean on all 6 touched files.

## Notes

- Builds on EVO-1235 [6.6] (the reviewed work). No dependency on the dormant-paths finding (#2).
- This repo has no model factories beyond `roles.rb` and does not include FactoryBot syntax (`csat_survey_spec` using `create(:inbox)` is itself non-running) — the new specs use plain `ActiveRecord.create!`, mirroring the request specs.

## Summary by Sourcery

Silence deprecated inline-content warnings for inbound messages and prevent sending blank auto-reply template messages for greeting and out-of-office flows.

Bug Fixes:
- Stop emitting deprecated inline-content WARN logs for inbound inbox messages that always carry content and have no template alternative.
- Avoid creating blank greeting auto-reply messages when both the referenced template and inline fallback yield empty content.
- Avoid creating blank out-of-office auto-reply messages when both the referenced template and inline fallback yield empty content.

Tests:
- Update inbound messages controller request spec to assert message creation without deprecated inline-content warnings.
- Add service specs for greeting and out-of-office template services covering blank-skipped replies, template render failures falling back to inline content, and no-template inline scenarios.

### DIFF
--- a/app/controllers/public/api/v1/inboxes/messages_controller.rb
+++ b/app/controllers/public/api/v1/inboxes/messages_controller.rb
@@ -1,6 +1,4 @@
 class Public::Api::V1::Inboxes::MessagesController < Public::Api::V1::InboxesController
-  include TemplateConsumerLogging
-
   before_action :set_message, only: [:update]
 
   def index
@@ -8,9 +6,6 @@ class Public::Api::V1::Inboxes::MessagesController < Public::Api::V1::InboxesCon
   end
 
   def create
-    # Coexistence (EVO-1235): flag external consumers still pushing inline body
-    # so they can be migrated to the templated send action.
-    warn_legacy_inline_content(action: 'messages#create', inbox_identifier: params[:inbox_id]) if permitted_params[:content].present?
     @message = @conversation.messages.new(message_params)
     build_attachment
     @message.save!
--- a/app/services/message_templates/template/greeting.rb
+++ b/app/services/message_templates/template/greeting.rb
@@ -4,8 +4,17 @@ class MessageTemplates::Template::Greeting
   pattr_initialize [:conversation!]
 
   def perform
+    content = greeting_content
+    # Skip silently when neither the template nor the inline fallback yields
+    # content, so a blank auto-reply is never delivered (EVO-1720 [6.11]).
+    return if content.blank?
+
     ActiveRecord::Base.transaction do
-      conversation.messages.create!(greeting_message_params)
+      conversation.messages.create!(
+        inbox_id: @conversation.inbox_id,
+        message_type: :template,
+        content: content
+      )
     end
   rescue StandardError => e
     EvolutionExceptionTracker.new(e, account: nil).capture_exception
@@ -17,15 +26,10 @@ class MessageTemplates::Template::Greeting
   delegate :contact, to: :conversation
   delegate :inbox, to: :message
 
-  def greeting_message_params
-    # Prefer a referenced MessageTemplate (EVO-1235); fall back to the inline string.
-    content = template_content_for(@conversation.inbox&.greeting_message_template_id) ||
-              @conversation.inbox&.greeting_message
-
-    {
-      inbox_id: @conversation.inbox_id,
-      message_type: :template,
-      content: content
-    }
+  # Prefer a referenced MessageTemplate (EVO-1235); fall back to the inline
+  # string. Resolved once so the blank guard adds no extra SendResolver lookup.
+  def greeting_content
+    template_content_for(@conversation.inbox&.greeting_message_template_id) ||
+      @conversation.inbox&.greeting_message
   end
 end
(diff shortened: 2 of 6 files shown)


dpaes (approved):
## Review — EVO-1720 [6.11] follow-up · APPROVED ✅

Non-blocking follow-up to the EVO-1235 [6.6] wire-call-sites review. All three findings verified against the code; no blockers.

**🟡 #1 [Medium] Inbound deprecated-content WARN — fixed.** `Public::Api::V1::Inboxes::MessagesController` no longer includes `TemplateConsumerLogging` nor calls `warn_legacy_inline_content`. The legitimate site (`OutboundMessagesController#outbound_content`) and the concern itself are untouched, and no dangling reference to the removed methods remains in the inbound controller. The request spec is correctly inverted to assert the WARN does NOT fire and the incoming message still persists with `message_type=incoming`.

**🟢 #3 [Low] Blank auto-reply — fixed** (greeting + out-of-office). `perform` resolves content once and `return if content.blank?` before `create!`. Traced the raise path: a `required` variable missing from the auto-reply's `{first_name,name,email}` makes `MessageTemplate#render_with_variables` raise `ArgumentError`; `TemplateContent#template_content_for` rescues to `nil`; the inline column is the fallback; a blank inline column is now skipped. Skipping (rather than persisting) the blank `:template` message keeps `HookExecutionService#first_message_from_contact?` — which counts `messages.template` — accurate, so a later valid greeting can still fire. That regression is genuinely closed.

**🟢 #2 [Low] Dormant accept-paths** (bot/macro/automation) intentionally untouched — documented in the commit.

**Spec quality:** the "render actually raises" guard tests are genuine — `extract_variables_from_content` (`before_save`) strips variables absent from `content`, so keeping `{{order_id}}` in the content is required to preserve the explicit `required:true`; the specs do exactly that. Setup mirrors the existing request spec, and `MessageTemplate.create!` without `language` is valid (column default `pt_BR`).

**Non-blocking notes:**
- This repo does not run RSpec in CI (Sourcery + contract only), so "14 examples, 0 failures" is self-reported — recommend a local `bundle exec rspec spec/services/message_templates/template/ spec/requests/public/api/v1/inboxes/messages_controller_spec.rb` before relying on it.
- Minor test-symmetry gap: `greeting_spec` has the "later valid greeting fires" case (AC6); `out_of_office_spec` has no equivalent. Skip behavior is identical, so non-blocking.
- Pre-existing (not introduced here): `delegate :inbox, to: :message` in `greeting.rb`/`out_of_office.rb` points to a non-existent `message` method (dead code) — worth removing while in the area.

Approving and merging. 🟢

PR #152, opened 2026-06-17
feat(inboxes): allow a global template as greeting/out-of-office message (EVO-1760)

daniloleonecarneiro:
## Summary

The inbox `greeting_message_template_id` / `out_of_office_message_template_id` columns (added in EVO-1235, read by the greeting / out-of-office send services) had no way in or out: inbox update did not permit them and neither `InboxSerializer` nor the inbox jbuilder returned them. So a global template could never be attached to an inbox greeting / OOO from the API.

- Permit both ids on inbox update (`inbox_attributes`).
- Expose both in `InboxSerializer` and `app/views/api/v1/models/_inbox.json.jbuilder`.

Pairs with the frontend PR that adds the picker.

## Test plan

- `spec/requests/api/v1/inboxes_template_refs_spec.rb` — **2 examples, 0 failures**: update persists + serializes both ids; the FE `"null"` sentinel clears them.
- Columns already exist (EVO-1235 migration) — no migration here.

## Linked issue
- EVO-1760

## Summary by Sourcery

Allow inbox greeting and out-of-office global templates to be set and exposed via the API.

New Features:
- Support configuring greeting and out-of-office message template IDs on inbox update and expose them in inbox API responses.

Tests:
- Add request specs covering persistence, serialization, and clearing of inbox greeting and out-of-office template IDs via the API.

### DIFF
--- a/app/controllers/api/v1/inboxes_controller.rb
+++ b/app/controllers/api/v1/inboxes_controller.rb
@@ -697,8 +697,10 @@ module Api
         end
 
         def inbox_attributes
-          [:name, :avatar, :display_name, :greeting_enabled, :greeting_message, :enable_email_collect, :csat_survey_enabled,
-           :enable_auto_assignment, :working_hours_enabled, :out_of_office_message, :timezone, :allow_messages_after_resolved,
+          [:name, :avatar, :display_name, :greeting_enabled, :greeting_message, :greeting_message_template_id,
+           :enable_email_collect, :csat_survey_enabled,
+           :enable_auto_assignment, :working_hours_enabled, :out_of_office_message, :out_of_office_message_template_id,
+           :timezone, :allow_messages_after_resolved,
            :lock_to_single_conversation, :sender_name_type, :business_name, :default_conversation_status,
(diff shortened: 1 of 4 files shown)


dpaes (approved):
**Approved** ✅ — backend half of EVO-1760 gap #1 (greeting/out-of-office), paired with FE #165.

Pure, correct wiring: `greeting_message_template_id` and `out_of_office_message_template_id` (UUID columns already created in EVO-1235) are added to the inbox permit list, the `as_json` serializer, and `_inbox.json.jbuilder` — so they round-trip permit → persist → serialize back to the FE. The request spec covers permit/persist/serialize and the clear path: both the `'null'` string sentinel (handled by the pre-existing CW-1567 normalization) and a real JSON `null` resolve to `nil`.

Verified the exposed columns feed a live send path (`HookExecutionService` → `Template::Greeting`/`OutOfOffice` → global-aware `SendResolver`), template preferred over inline free text, safe fallback on miss/render error — so this is not an inert toggle.

Note: this repo's CI runs Sourcery/staleness/contract only (no rspec), so the green spec is dev-reported — worth a local `bundle exec rspec spec/requests/api/v1/inboxes_template_refs_spec.rb` before relying on it.

PR #156, opened 2026-06-22
feat(crm): conversations history import via DataImport (EVO-1557)

marcelogorutuba:
## Summary
- Extends `DataImport` to accept `data_type='conversations'` for migrating chat history from other CRMs.
- New service `DataImport::ConversationManager`: denormalized CSV (1 row per message) → contact lookup (identifier → phone → email), idempotent upsert by `conversation.identifier` and `message.source_id`, lazy-creates one `Imported History` Channel::Api inbox, imported conversations land as `resolved`.
- New endpoint `POST /api/v1/conversations/import` (multipart): `202 + { data_import_id }` on success; `422` on missing file, > 50k rows, > 50MB, or malformed CSV.
- `DataImportJob` now dispatches by `data_type`; the existing contacts flow is unchanged.
- rack-attack throttle `api/v1/conversations/import` (default 5/min/key, env `RATE_LIMIT_CONVERSATIONS_IMPORT`).

## Acceptance Criteria Coverage
All 11 ACs from EVO-1557 are implemented and covered by automated tests:

| AC | Coverage |
|---|---|
| AC1 — validator accepts `conversations` | `conversation_manager_spec` |
| AC2 — 202 + `data_import_id` | `conversations_import_spec` |
| AC3 — Job processes via `ConversationManager` | `data_import_job_spec` |
| AC4 — required CSV columns | `conversation_manager_spec` (missing header) |
| AC5 — orphan contact → row fails, processing continues | `conversation_manager_spec` |
| AC6 — unmapped sender → user_id=NULL, sender_name preserved | `conversation_manager_spec` |
| AC7 — idempotent re-import | `conversation_manager_spec` |
| AC8 — `message_type != 'text'` → `[mídia: {type}]` | `conversation_manager_spec` |
| AC9 — CSV > 50k rows → 422 | `conversations_import_spec` |
| AC10 — structured report (`total_rows`, `success_count`, `error_count`, `errors[]`) | `data_import_job_spec` |
| AC11 — specs for orphan contact, unmapped sender, idempotent, malformed | `conversation_manager_spec` |

## Security
- Permission `conversations.import` required.
- rack-attack throttle (5 req/min/key, token-hashed discriminator falling back to IP).
- Upfront byte limit (50MB) before ActiveStorage attach to prevent DoS.
- Upfront row count (streaming `CSV.foreach`, breaks at limit+1) to short-circuit oversized files without loading 50k rows into heap.
- `CSVSafe.generate` used for `failed_records` output (CSV-injection-safe).
- Empty `contact_identifier` is rejected; phone fallback skipped when value strips to empty; email fallback skipped unless value contains `@` — prevents `find_by(phone_number: nil)` false-positive matches.

## Test plan
- `evo-ai-crm-community: bundle exec rspec spec/services/data_import/conversation_manager_spec.rb spec/jobs/data_import_job_spec.rb spec/requests/api/v1/conversations_import_spec.rb` — 19 examples, 0 failures
- Manual smoke (`rails runner`) end-to-end confirmed: 3/4 rows imported (1 orphan rejected), failed_records CSV attached, idempotent re-import keeps the same conversation id and message count

## Changed Files
- `app/models/data_import.rb`
- `app/services/data_import/conversation_manager.rb` (new)
- `app/jobs/data_import_job.rb`
- `config/routes.rb`
- `config/initializers/rack_attack.rb`
- `config/locales/en.yml`
- `config/locales/pt_BR.yml`
- `spec/services/data_import/conversation_manager_spec.rb` (new)
- `spec/jobs/data_import_job_spec.rb` (new)
- `spec/requests/api/v1/conversations_import_spec.rb` (new)

## Known limitations (deferred LOW findings from adversarial review)
- No completion notification mailer for conversations success path (contacts flow has `contact_import_complete`; conversations is silent). Deferred to the UI story.
- `errors.data_import.data_type.invalid` i18n key shared between both data types (no per-type customization yet).
- `count_csv_rows` returns `[0, nil]` if uploaded file lacks `.path` (Rails multipart always has it; defensive fail-open could be tightened).
- Logging in the conversations flow is lighter than the contacts flow.
- Inbox lookup by `display_name` could collide with a manually-created inbox of the same name (no marker on `additional_attributes`).
- Idempotency spec does not assert that `conversation.status` stays `resolved` across re-imports.
- No spec for malformed `sent_at` timestamp specifically.
- `failed_records` CSV does not include a `_row_number` column.

## Linked Issue
- EVO-1557

## Summary by Sourcery

Add support for importing conversation history via the DataImport pipeline and expose a throttled API endpoint for CSV-based conversation imports.

New Features:
- Introduce a DataImport data_type for conversations to process imported conversation history CSV files.
- Add a conversations import endpoint that validates CSV size and structure before queuing a background data import job.
- Implement a ConversationManager service to upsert conversations and messages from import rows, including idempotency and automatic resolution of imported conversations.

Enhancements:
- Extend DataImportJob to dispatch processing based on data_type and to generate structured reports and failed-records CSVs for conversation imports.
- Add localized error messages for conversations import validation failures in English and Brazilian Portuguese.

Tests:
- Add service, job, and request specs covering conversation imports, CSV validation, and reporting for the new data_type.

### DIFF
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -21,10 +21,14 @@ class Api::V1::ConversationsController < Api::V1::BaseController
     transcript: 'conversations.export',
     email_team: 'conversations.update',
     available_for_pipeline: 'conversations.read',
-    unread_count: 'conversations.read'
+    unread_count: 'conversations.read',
+    import: 'conversations.import'
   })
 
-  before_action :conversation, except: [:index, :meta, :search, :create, :filter, :unread_count]
+  CONVERSATIONS_IMPORT_ROW_LIMIT = 50_000
+  CONVERSATIONS_IMPORT_MAX_BYTES = 50 * 1024 * 1024
+
+  before_action :conversation, except: [:index, :meta, :search, :create, :filter, :unread_count, :import]
   before_action :inbox, :contact, :contact_inbox, only: [:create]
 
   ATTACHMENT_RESULTS_PER_PAGE = 100
@@ -53,10 +57,61 @@ class Api::V1::ConversationsController < Api::V1::BaseController
   def meta
     result = conversation_finder.perform
     @conversations_count = result[:count]
-    
+
     success_response(
       data: { count: @conversations_count },
       message: 'Conversation metadata retrieved successfully'
+    )
+  end
+
+  def import
+    if params[:import_file].blank?
+      return error_response(
+        ApiErrorCodes::MISSING_REQUIRED_FIELD,
+        I18n.t('errors.conversations.import.missing_file'),
+        details: { field: 'import_file', message: 'is required' },
+        status: :unprocessable_entity
+      )
+    end
+
+    file_size = params[:import_file].respond_to?(:size) ? params[:import_file].size : 0
+    if file_size > CONVERSATIONS_IMPORT_MAX_BYTES
+      return error_response(
+        ApiErrorCodes::INVALID_PARAMETER,
+        I18n.t('errors.conversations.import.too_large_bytes', limit: CONVERSATIONS_IMPORT_MAX_BYTES),
+        details: { byte_size: file_size, limit: CONVERSATIONS_IMPORT_MAX_BYTES },
+        status: :unprocessable_entity
+      )
+    end
+
+    row_count, malformed_error = count_csv_rows(params[:import_file])
+    if malformed_error
+      return error_response(
+        ApiErrorCodes::INVALID_PARAMETER,
+        I18n.t('errors.conversations.import.invalid_csv', error: malformed_error),
+        status: :unprocessable_entity
+      )
+    end
+
+    if row_count > CONVERSATIONS_IMPORT_ROW_LIMIT
+      return error_response(
+        ApiErrorCodes::INVALID_PARAMETER,
+        I18n.t('errors.conversations.import.too_large', limit: CONVERSATIONS_IMPORT_ROW_LIMIT),
+        details: { row_count: row_count, limit: CONVERSATIONS_IMPORT_ROW_LIMIT },
+        status: :unprocessable_entity
+      )
+    end
+
+    data_import = ActiveRecord::Base.transaction do
+      import = DataImport.create!(data_type: 'conversations')
+      import.import_file.attach(params[:import_file])
+      import
+    end
+
+    success_response(
+      data: { data_import_id: data_import.id },
+      message: 'Conversations import accepted',
+      status: :accepted
     )
   end
 
@@ -386,6 +441,25 @@ class Api::V1::ConversationsController < Api::V1::BaseController
   end
 
   private
+
+  def count_csv_rows(uploaded_file)
+    path = uploaded_file.respond_to?(:path) ? uploaded_file.path : nil
+    return [0, nil] if path.nil?
+
+    cap = CONVERSATIONS_IMPORT_ROW_LIMIT + 1
+    data_rows = 0
+    begin
+      CSV.foreach(path, headers: true) do |_row|
+        data_rows += 1
+        break if data_rows >= cap
+      end
+    rescue CSV::MalformedCSVError => e
+      return [0, e.message]
+    rescue Errno::ENOENT
+      return [0, 'file is not readable']
+    end
+    [data_rows, nil]
+  end
 
   def update_custom_attribute(attribute_key, value, success_message)
     custom_attributes = (@conversation.custom_attributes || {}).merge(attribute_key => value)
--- a/app/jobs/data_import_job.rb
+++ b/app/jobs/data_import_job.rb
@@ -32,6 +32,11 @@ class DataImportJob < ApplicationJob
 
   def process_conversations_import
     @data_import.update!(status: :processing)
+    started_at = Time.current
+    Rails.logger.info(
+      "[DataImport::Conversation] start data_import_id=#{@data_import.id}"
+    )
+
     manager = DataImport::ConversationManager.new(@data_import)
     report = manager.process
 
@@ -40,6 +45,13 @@ class DataImportJob < ApplicationJob
       total_records: report['total_rows'],
       processed_records: report['success_count'],
       processing_errors: report.to_json
+    )
+
+    duration = (Time.current - started_at).round(2)
+    Rails.logger.info(
+      "[DataImport::Conversation] done data_import_id=#{@data_import.id} " \
+      "rows_total=#{report['total_rows']} imported=#{report['success_count']} " \
+      "failed=#{report['error_count']} duration=#{duration}s"
     )
 
     save_conversations_failed_records(manager)
--- a/app/jobs/data_import_job.rb
+++ b/app/jobs/data_import_job.rb
@@ -7,11 +7,16 @@ class DataImportJob < ApplicationJob
 
   def perform(data_import)
     @data_import = data_import
-    @contact_manager = DataImport::ContactManager.new
-    Rails.logger.info "📊 DataImportJob: Starting import for data_import_id=#{@data_import.id}"
+    Rails.logger.info "📊 DataImportJob: Starting import for data_import_id=#{@data_import.id} data_type=#{@data_import.data_type}"
     begin
-      process_import_file
-      send_import_notification_to_admin
+      case @data_import.data_type
+      when 'conversations'
+        process_conversations_import
+      else
+        @contact_manager = DataImport::ContactManager.new
+        process_import_file
+        send_import_notification_to_admin
+      end
       Rails.logger.info "📊 DataImportJob: Import completed for data_import_id=#{@data_import.id}"
     rescue CSV::MalformedCSVError => e
       Rails.logger.error "📊 DataImportJob: CSV error for data_import_id=#{@data_import.id}: #{e.message}"
@@ -24,6 +29,40 @@ class DataImportJob < ApplicationJob
   end
 
   private
+
+  def process_conversations_import
+    @data_import.update!(status: :processing)
+    manager = DataImport::ConversationManager.new(@data_import)
+    report = manager.process
+
+    @data_import.update!(
+      status: :completed,
+      total_records: report['total_rows'],
+      processed_records: report['success_count'],
+      processing_errors: report.to_json
+    )
+
+    save_conversations_failed_records(manager)
+  end
+
+  def save_conversations_failed_records(manager)
+    rejected = manager.rejected_rows
+    return if rejected.blank?
+
+    headers = (manager.csv_headers || rejected.first.keys.reject { |k| k.start_with?('_') || k == 'errors' }) + ['errors']
+    csv_data = CSVSafe.generate do |csv|
+      csv << headers
+      rejected.each do |record|
+        csv << headers.map { |h| record[h] }
+      end
+    end
+
+    @data_import.failed_records.attach(
+      io: StringIO.new(csv_data),
+      filename: "#{Time.zone.today.strftime('%Y%m%d')}_conversations_failed.csv",
+      content_type: 'text/csv'
+    )
+  end
 
   def process_import_file
     @data_import.update!(status: :processing)
@@ -198,7 +237,7 @@ class DataImportJob < ApplicationJob
   def handle_csv_error(error) # rubocop:disable Lint/UnusedMethodArgument
     Rails.logger.error "📊 DataImportJob: Handling CSV error - marking as failed"
     @data_import.update!(status: :failed)
-    send_import_failed_notification_to_admin
+    send_import_failed_notification_to_admin if @data_import.data_type == 'contacts'
   end
 
   def send_import_notification_to_admin
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -68,6 +68,7 @@ class Conversation < ApplicationRecord
 
   enum status: { open: 0, resolved: 1, pending: 2, snoozed: 3 }
   enum priority: { low: 0, medium: 1, high: 2, urgent: 3 }
+  enum source: { live: 0, imported: 1 }
 
   scope :unassigned, -> { where(assignee_id: nil) }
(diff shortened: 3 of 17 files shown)


dpaes (changes requested):
## Code review — Changes requested

Solid implementation overall: clean parity with `DataImport::ContactManager`, the adversarial M1–M4 fixes are well done, idempotency is well designed, and all 11 ACs are functionally covered. Requesting changes on the items below before merge. Severity legend: 🔴 merge gate · 🟠 high (needs decision) · 🟡 medium · 🟢 low.

### 🔴 B1 — Branch conflicts with `develop` (merge gate)
Single, trivial conflict in `config/initializers/rack_attack.rb`: both this PR and EVO-1735 (already on `develop`) inserted a new `throttle(...)` block at the same spot, right before the trailing `## ---- ##` marker. Resolution = keep both blocks. `routes.rb` auto-merges. Please rebase onto current `develop` and resolve.

### 🟠 H1 — Import runs the full *live* message/conversation callback chain (needs product decision)
Records are created with `find_or_create_by!` / `conversation.messages.create!` (full callbacks), not `insert_all`/`skip_callbacks`. So every imported message fires the present-tense pipeline:

- **`notify_via_mail`** — in this fork `email_notifiable_api_channel?` returns `inbox.api?` with **no gate**, so every imported `outgoing` message for a contact **with an email** enqueues `ConversationReplyEmailWorker` → a **real email is sent to the contact**. On a historical backfill this spams customers.
- **`dispatch_create_events` (MESSAGE_CREATED)** → fires **automation rules, agent bots, and reporting_events** (skews analytics) per message.
- **`publish_message_created` / `publish_conversation_created`** (Wisper) → feeds **EvoFlow**, can trigger **journeys** on historical data.
- **`assign_to_default_pipeline`** (Conversation callback) → **every imported conversation lands in the default pipeline**, flooding the board.
- **`reopen_conversation`** → an incoming message re-opens the resolved conversation (`open!`); `resolve_touched_conversations` later fixes the final state via `update_all` (no callbacks), but the intermediate open/update events still fire.

Confirmed mitigation: `SendReplyJob` is a **no-op for `Channel::Api`** (no `SendOnApiService` in the service map), so the import does **not** re-deliver messages through the channel. The effects above still occur. Specs don't assert the *absence* of these side effects (in the test env jobs only queue, and no pipeline/automation is configured, so they pass without surfacing it).

Recommendation: suppress callbacks on the backfill path (bulk `insert_all`, a `Current`-scoped "imported" guard the callbacks check, or controlled `skip_callbacks`). If the team chooses to accept the effects as-is, please get explicit product sign-off and document it.

### 🟡 M1 — New `conversations.import` permission must be registered in evo-auth-service
The real authz gate is remote (`require_permissions` → `EvoAuthService.check_user_permission`); the local `User#has_permission?` is a no-op. `conversations.import` is a brand-new key. If it isn't added to the auth-service permission catalog (resource + action) and re-seeded, the grant is dropped silently and **nobody can import in production (403)** — including admins. Specs stub the permission check, so they pass regardless. Please confirm/seed it.

### 🟡 M2 — No per-row transaction → orphan empty conversations
For each row the flow does (1) create/find conversation → (2) create message. These aren't wrapped per-row. If step (2) raises after step (1) created a new conversation (e.g. the M1 `RecordInvalid` case for content > 150KB), the conversation persists with **zero messages**, still marked `resolved` and pipeline-assigned. Wrap `import_row` in a transaction so a failed row rolls back its conversation/contact_inbox.

### 🟡 M3 — AC2 mentions `column_mapping JSON`, not implemented (fixed schema)
The endpoint takes only the file and requires exact column names. This matches the contacts-import precedent (also no `column_mapping`) and the card's "CSV format" section, so fixed schema is likely intended — just confirm with the PM so AC2 is formally closed.

### 🟢 Low (non-blocking)
- **L1** — Re-import of rows **without** `message_external_id` duplicates messages (dedup is only by `source_id`). Consistent with AC7 wording, but worth a UX note.
- **L2** — `Conversation.find_or_create_by!(identifier:)` uses a column with **no unique index** and no scope; a collision with a pre-existing conversation sharing the `identifier` string would attach imported messages to it. Low probability.
- **L3** — `ensure_imported_history_inbox` / `ensure_conversation` have no lock → concurrent imports could duplicate the "Imported History" inbox. Serial jobs make this unlikely.
- **L4** — `count_csv_rows` returns `[0, nil]` when `.path` is nil, bypassing the row-limit and malformed checks (already documented; `ActionDispatch::Http::UploadedFile` always has a path in practice).
- **L5** — `failed_records` CSV omits `_row_number` (documented); `errors[]` JSON capped at 1000 (reasonable).
- **L7** — encoding round-trip via `UTF-16le` is convoluted; `String#scrub` would be cleaner (cosmetic).

### Notes
- AC2 returns `202 + data_import_id` — better than the contacts import (200, no id). 👍
- Single-tenant assumption is correct: `conversations`/`messages`/`inboxes` have no `account_id` in this fork → no cross-account risk.
- Reminder: `evo-ai-crm-community` doesn't run RSpec in CI, so the "19 examples, 0 failures" is self-reported — please share a local `bundle exec rspec` run.

PR stays open. Happy to re-review after the rebase + H1 decision.

dpaes (changes requested):
## Code review — round 2 · CHANGES REQUESTED · PR #156

Good progress: **M2** (per-row transaction) and the **Message side of H1** are properly fixed, and the `source` enum approach is clean. Requesting changes on the points below — two merge gates remain (one worse than round 1) and H1 is only half-resolved. PR stays **open**.

Legend: 🔴 merge gate · 🟠 high (needs change/decision) · 🟡 medium · 🟢 low · ℹ️ info.

### 🔴 B1/B2 — `db/schema.rb` is corrupted (regenerated against a shared multi-service DB)

The branch is still **CONFLICTING** vs `develop` — the rack-attack conflict from round 1 is resolved, but `schema.rb` was regenerated against a Postgres DB that also had **other services' tables applied**, producing three classes of damage, all absent from `develop`:

1. **~11 alien tables** dumped in: `events`, `sessions`, `user_states`, `plans`, `plan_features`, `features`, `evo_core_agents`, `evo_core_api_keys`, `evo_core_folders`, `evo_ai_agent_processor_sessions`, `evo_enterprise_alembic_version` (these belong to evo-ai-core-service / auth-service).
2. **Mutation of an existing table**: `user_tours` becomes `id: :uuid` + `completed_at null: false` on the branch — **contradicting its own migration** `20260414120000_create_user_tours.rb` (bigint PK / nullable on both branches). A `db:schema:load` would create the wrong table.
3. **17 phantom foreign keys** declared by no migration (`facebook_comment_moderations→users`, `pipeline_tasks→users`, `scheduled_actions→users`, …) — schema-load vs schema-migrate parity break.

This is the **root cause of the merge conflict** and would contaminate the CRM's canonical schema. **Fix:** regenerate `schema.rb` from a clean DB with only the CRM migrations (or resolve the conflict keeping just `t.integer "source", default: 0, null: false` on `messages` + the version bump), then rebase on `develop`. The only legitimate schema change this PR needs is the `messages.source` column.

### 🟠 H1 — only the Message side is gated; Conversation callbacks still fire live

The Message gating is solid (enum `Message#source` + `unless: :imported?` covers flooding, `execute_after_create_commit_callbacks`, `publish_message_created`, `sync_message_event`; serializers don't leak `source`). 👍

But `Conversation.find_or_create_by!` in `ConversationManager#ensure_conversation` fires **4 ungated `after_create_commit` callbacks** per imported conversation:

- `notify_conversation_creation` → `dispatcher_dispatch(CONVERSATION_CREATED)` → **conversation_created automation rules** (`AutomationRuleListener#conversation_created` does not filter imports), **reporting_events** (distorts analytics), **outbound webhooks**;
- `publish_conversation_created` → Wisper → **EvoFlow** → journeys over historical data;
- `assign_to_default_pipeline` → imported conversation lands in `Pipeline.default` (or the contact's active pipeline) → **floods the board**;
- `load_attributes_created_by_db_triggers` (benign — one extra SELECT).

On a realistic backfill spanning thousands of conversations this triggers retroactive live side-effects. Note the `additional_attributes['imported']` marker the service writes is **read nowhere** in app code — it's decorative. **Required fix:** gate the three problematic Conversation callbacks on the import path (same approach as Message). (Product confirmed this is a correctness fix, not an accepted trade-off.)

### 🟡 M1 — `conversations.import` permission not registered (still open from round 1)

`conversations.import` is not in the auth-service `ResourceActionsConfig` (which has `import` only for agents/contacts/templates) nor in the `rbac.rb` seeds. `check_permission!` has no admin bypass — only a **service token** passes. So user/UI import = **403 in prod, including admins**. Add the resource+action to the catalog, grant it to the appropriate role, and re-seed.

### 🟡 `sent_at` parsing

- **Strict parser** — `Time.iso8601` rejects the most common export formats (space-separated `"YYYY-MM-DD HH:MM:SS"`, date-only, unix-epoch). An entire export can fail row-by-row and import 0 messages while the job reports `status=completed`. Failures do surface in the AC10 report (recoverable), but there's no documented format contract and no negative-format spec. Given the denormalized model, one bad format wipes the whole conversation. Suggest accepting common formats (or `Time.zone.parse` with a documented contract).
- **Naive timestamps interpreted in server-local time** — `Time.iso8601("…T10:30:00")` without an offset is parsed as process-local time; with `created_at` stored as `timestamp without time zone` (UTC), this shifts the wall-clock by the server offset (e.g. +3h on a BRT box) and can reorder the timeline. Default naive values to UTC (`Time.find_zone('UTC').parse` / `in_time_zone`).

### 🟢 Low
- **L1 (round 1)** — re-import of rows **without** `message_external_id` duplicates messages (dedup only by `source_id`); AC7 breaks for exports lacking a stable per-message id. Consider a synthetic dedup key (conversation + sent_at + direction + content hash) or rejecting blank ids.
- `last_activity_at` stays at import wall-clock (never derived from `sent_at`), so imported historical conversations sort to the top of the agent inbox as if just active. **(new)**
- **L2 (round 1)** — `find_or_create_by!(identifier:)` is unscoped/non-unique; realistic collision surface is narrow (only existing Slack-thread timestamps), but the blind force-resolve-via-`update_all` of a matched conversation is worth scoping (e.g. require `additional_attributes->>'imported'` or scope by inbox).
- **L3 (round 1)** — `ensure_imported_history_inbox` find-or-create race (no unique constraint on inbox `display_name`); narrow window (first/concurrent first import), pollution-only.
- `belongs_to :conversation, touch: true` issues ~one redundant conversation UPDATE + an (inert, zero-subscriber) `:conversation_updated` Wisper publish per imported message → 50k extra UPDATEs on a 50k-row import. **(new, low)**
- Spec `does NOT fire message_created listeners`: 2 of 5 assertions pass vacuously (async listeners never run under `queue_adapter=:test`). The gating itself is correct and is proven by the SendReplyJob/SessionSync/ActionCable assertions, but the Webhook/AutomationRule assertions prove nothing.

### ℹ️ Notes
- New migration uses `ActiveRecord::Migration[7.0]` vs `[7.1]` schema — cosmetic (16 pre-existing `[7.0]` migrations); no behavior change for a scalar `add_column`.
- Reminder: crm-community does **not** run RSpec in CI → "27/29 passing" is self-reported. Please attach `bundle exec rspec` output, and confirm the 2 cited failures (`conversations_import_spec:124`, `message_spec:9`) are pre-existing (a `git stash` check isn't reproducible here).
- AC2 `column_mapping` is still a fixed schema (consistent with the contacts-import precedent) — fine to keep, just confirm with PM to formally close AC2.

Moving the card back to **Todo**. Re-review after: schema.rb regenerated + rebased, Conversation callbacks gated, `conversations.import` seeded.

dpaes:
**Review for EVO-1557 (backend)** — card kept In Review; this PR stays open pending rebase.

All 11 ACs are functionally met: multipart `POST /conversations/import` returns 202 + DataImport.id, the 50k limit is enforced by streaming **before** enqueue, missing contact fails the row and continues, unmapped sender preserves `sender_name`, idempotent upsert by `message_external_id`, non-text content becomes `[mídia: {type}]`, the report carries `total_rows/success_count/error_count/errors[]`, and specs cover the five cases. Live side-effects are correctly suppressed via the new `source: imported` enum. Single-tenant repo, so the cross-tenant lookup trap doesn't apply.

**Blocking state:** the branch is CONFLICTING with develop on `db/schema.rb` (version line: develop `2026_06_23_110717` vs this PR `2026_06_23_120000`) — mechanical, not logical. Please rebase onto `origin/develop` and regenerate `schema.rb`. (Pinged you on Slack.)

**[medium]** Re-import dedup is app-level only: rows with a blank `message_external_id` re-create on every run (duplicates), and with no unique index on `conversations.identifier` / `messages.source_id` plus non-atomic `find_or_create_by!`, concurrent imports of the same conversation can duplicate. Document the limitation or add a partial unique index + rely on `RecordNotUnique`.

Note: the RBAC sibling (auth-service #49, `conversations.import` permission) is **already merged** to develop, so the permission is in place. CI doesn't run RSpec here — please paste local `bundle exec rspec` output before merge.

dpaes (approved):
Reviewed against EVO-1557. Conversations DataImport is wired end-to-end: the validator now accepts `data_type='conversations'`, `POST /api/v1/conversations/import` returns 202 + DataImport.id, `DataImportJob` dispatches to `DataImport::ConversationManager`, and the manager performs an idempotent upsert (conversation by external id; messages skipped by existing `message_external_id`, new ones appended). All 11 ACs verified against real-path specs (`conversation_manager_spec`, `conversations_import_spec`, job spec) — not mocked-away. Live-message side-effects (notify/publish/pipeline/flooding/touch) are correctly suppressed for imported records via the `source` enum. The >50k guard runs before the job is enqueued (422), and the final report carries `total_rows/success_count/error_count/errors[]`.

One documented limitation (non-blocking): rows without `message_external_id` are not idempotent and will duplicate on re-import — consistent with AC7 (which scopes idempotency to rows that carry the id); worth surfacing in the future import-template/UI story. Note: CRM CI does not run RSpec, so the green spec run is self-reported; specs are real-path and credible. LGTM — approving.

PR #158, opened 2026-06-22
feat(crm): update_custom_attribute automation action (EVO-1751)

nickoliveira23:
## Summary
- New `update_custom_attribute` automation action: sets/updates a custom attribute value, routing by the definition's `attribute_model` so one action covers **contact**, **conversation** and **pipeline-item** attributes.
- Added to the model action whitelist; dispatched via `ActionService#send`; enabled for contact-triggered rules in `ContactActionService`.
- `attribute_key` is only unique per `attribute_model`, so `action_params` carry the model to disambiguate the lookup and the write target. Values are stored verbatim in the JSONB column (string contract), cast at read/query time — consistent with the journey/CRM convention.

## Security
- Single-tenant (no `account_id`); the handler writes only to records already in the rule's scope (`@conversation` / its contact / its pipeline_item).
- Validates that `(attribute_key, attribute_model)` exists before writing; no-op otherwise. JSONB merge, no raw SQL. `Current.executed_by = @rule` on the contact path (loop guard + dirty-tracking).

## Test plan
- `crm: bundle exec rspec spec/services/automation_rules/action_service_custom_attribute_spec.rb spec/services/automation_rules/contact_action_service_spec.rb`
- `crm: bundle exec rubocop app/models/automation_rule.rb app/services/automation_rules/conversation_action_handlers.rb app/services/automation_rules/contact_action_service.rb`
- Manual smoke: rule on `conversation_updated` (no conditions) → action wrote `onboarding_status` to the conversation's contact (verified in DB).

> Note: specs were authored + syntax/rubocop-validated but not executed locally — on this branch the `test` DB resolves to the shared dev DB (`maintain_test_schema!` risk); they run in CI.

## Changed Files
- `app/models/automation_rule.rb`
- `app/services/automation_rules/conversation_action_handlers.rb`
- `app/services/automation_rules/contact_action_service.rb`
- `spec/services/automation_rules/action_service_custom_attribute_spec.rb`
- `spec/services/automation_rules/contact_action_service_spec.rb`

## Related PRs
- frontend: _(see linked frontend PR for the builder UI)_

## Linked Issue
- EVO-1751

## Summary by Sourcery

Add a new automation action to update custom attributes across conversations, contacts, and pipeline items, and wire it into the automation rule and contact action handling.

New Features:
- Introduce an update_custom_attribute automation action that can modify conversation, contact, and pipeline-item custom attributes based on a provided model and key.

Enhancements:
- Extend automation rule action whitelisting and contact-native actions to include the new custom-attribute update action.

Tests:
- Add service specs to verify custom attribute updates via contact-triggered automation, including contact-only scenarios without a conversation context.

### DIFF
--- a/app/models/automation_rule.rb
+++ b/app/models/automation_rule.rb
@@ -47,7 +47,8 @@ class AutomationRule < ApplicationRecord
   def actions_attributes
     %w[send_message send_canned_response send_template add_label remove_label send_email_to_team assign_team assign_agent
        send_webhook_event mute_conversation send_attachment change_status resolve_conversation snooze_conversation
-       change_priority send_email_transcript assign_to_pipeline update_pipeline_stage create_pipeline_task].freeze
+       change_priority send_email_transcript assign_to_pipeline update_pipeline_stage create_pipeline_task
+       update_custom_attribute].freeze
   end
 
   def file_base_data
--- a/app/services/automation_rules/contact_action_service.rb
+++ b/app/services/automation_rules/contact_action_service.rb
@@ -42,6 +42,7 @@ class AutomationRules::ContactActionService
     action_params = action[:action_params]
 
     return record_skip(action_name) unless CONTACT_NATIVE_ACTIONS.include?(action_name)
+    return record_skip(action_name) if action_name == 'update_custom_attribute' && !contact_scoped_attribute?(action_params)
 
     dispatch_native_action(action_name, action_params)
     @recorder&.add_step("Action: #{action_name}", level: 'success', data: { params: action_params })
@@ -60,6 +61,14 @@ class AutomationRules::ContactActionService
         reason: 'requires a conversation; contact trigger has no conversation in scope'
       }
     )
+  end
+
+  # update_custom_attribute is contact-native ONLY for contact_attribute targets;
+  # conversation/pipeline_item attributes need a conversation, so on a contact
+  # trigger they would be a silent no-op — record them as skipped instead.
+  def contact_scoped_attribute?(action_params)
+    raw = Array(action_params).first
+    raw.is_a?(Hash) && raw.with_indifferent_access[:custom_attribute_model].to_s == 'contact_attribute'
   end
 
   # The action_name is constrained to CONTACT_NATIVE_ACTIONS before we get here,
--- a/app/services/automation_rules/contact_action_service.rb
+++ b/app/services/automation_rules/contact_action_service.rb
@@ -18,7 +18,7 @@ class AutomationRules::ContactActionService
   include AutomationRules::ConversationActionHandlers
 
   # Actions that operate on the contact itself and need no conversation.
-  CONTACT_NATIVE_ACTIONS = %w[send_webhook_event add_label remove_label].freeze
+  CONTACT_NATIVE_ACTIONS = %w[send_webhook_event add_label remove_label update_custom_attribute].freeze
 
   def initialize(rule, contact, recorder: nil)
     @rule = rule
@@ -69,6 +69,7 @@ class AutomationRules::ContactActionService
     when 'send_webhook_event' then send_webhook_event(action_params)
     when 'add_label' then add_label(action_params)
     when 'remove_label' then remove_label(action_params)
+    when 'update_custom_attribute' then update_custom_attribute(action_params)
     end
   end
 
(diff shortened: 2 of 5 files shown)


dpaes:
## Review — Changes requested 🔴 (EVO-1751, backend half)

Reviewed together with frontend #172 (one feature). The cross-repo wire contract is **clean** (action type string, `custom_attribute_{key,model,value}` keys, 3-model vocabulary, string value — all match #172). Security is **solid**: fixed model routing, `CustomAttributeDefinition.exists?` guard before writing, writes only to dedicated JSONB columns — no mass-assignment, no arbitrary model/column writes.

Sending back on test verification + minor observability:

**MEDIUM — specs not executed, CI doesn't run RSpec:**
You noted the specs were authored but not run locally (test env resolves to the shared dev DB) and CRM CI runs no RSpec, so the write path is effectively **unverified** — notably the `attribute_model` enum-int ↔ string mapping the lookup depends on. Please run `bundle exec rspec spec/services/automation_rules/action_service_custom_attribute_spec.rb spec/services/automation_rules/contact_action_service_spec.rb` locally and paste the green output.

LOW:
- A contact-triggered rule targeting a `conversation_attribute`/`pipeline_item_attribute` passes the `exists?` check but early-returns on the write (`return unless @conversation`) while the recorder still logs `level: 'success'` — misleading no-op. Record an explicit skip, or restrict selectable models to the trigger type on the frontend.
- `pipeline_items.first` is non-deterministic with >1 item (consistent with the existing `create_pipeline_task`, so parity — optional to make deterministic).

Leaving the PR open.

dpaes (approved):
## Re-review — APPROVED ✅

Evaluated as a pair: **crm #158 (AC2/backend)** + **frontend #172 (AC1/builder)**. Both base `develop`, CLEAN/MERGEABLE.

All round-1 findings are resolved and independently verified against the canonical code (direct read + a 3-agent adversarial panel, all `refuted: false`, high confidence):

- **H1 (HIGH)** — `update_custom_attribute` added to `BACKEND_ACTION_NAMES`; the front↔backend parity guard is green (vitest 8/8). ✅
- **M1 (MED)** — checkbox is now coerced to a **real boolean at the backend write boundary** (`cast_custom_attribute_value` by `attribute_display_type`). Verified correct: every canonical reader uses `Boolean(rawValue)` truthiness (`CustomAttributesForm.tsx:364/561`, `PipelineCustomAttributesForm.tsx:498`, chat header/sidebar), so a real boolean renders correctly for both `true` and `false` — and it actually fixes the latent `Boolean('false') === true` trap. Keeps the string wire contract; non-checkbox types stay verbatim (cast at read/query time, matching `filter_service` `::numeric`/`::date`). ✅
- **M2 (MED)** — both specs run green on an isolated test DB (15 examples, 0 failures). Confirmed the specs exercise the real dispatch (`send(action_name)` gated by `AutomationRule#actions_attributes`, no mocking). ✅
- **L2 (LOW)** — a contact-triggered `update_custom_attribute` targeting a conversation/pipeline attribute now records an explicit `warn` skip instead of a misleading `success` on a silent no-op. ✅

Deferred with sound justification: **L1** (front↔back `UPDATABLE_ATTRIBUTE_MODELS` drift — a real guard needs cross-repo CI tooling that doesn't exist; verified manually), **L3** (`pipeline_items.first` — consistent with the existing `create_pipeline_task`).

Additional verification: cross-repo wire contract matches char-for-char (`custom_attribute_{key,model,value}`, 3-model vocabulary, string value both sides); backend write is safe (definition-gated lookup, fixed JSONB columns, no mass-assignment); the automation-loop guard covers all three write paths (`Current.executed_by` set globally + `performed_by_automation?` / `contact.rb` guard / pipeline stage-change gate); AC1 populates end-to-end.

### Non-blocking notes (informational, not gating)
- Type heterogeneity: checkbox stored as a native boolean while number/currency/date stay as wire strings — internally consistent (query-time casts), noted for any future native-type reader.
- Pre-existing: a `pipeline_item_attribute` whose `attribute_key` is literally `services`/`currency` would merge over the reserved structured deal-value in `custom_fields` (also reachable via the existing sidebar; `attribute_must_not_conflict` doesn't block those keys). Not introduced here.

Squash-merging both and deleting the `feat/EVO-1751` branches. Submodule pointer bumps in the umbrella are handled in the release batch, not here.

PR #160, opened 2026-06-23
fix(crm): return structured field errors from products endpoints (EVO-1783)

nickoliveira23:
## Summary
Companion backend change for EVO-1783 (frontend PR #174 review). The products and variants controllers called `error_response` with keyword arguments (`code:`, `message:`) while the helper signature is positional `error_response(code, message, details:, status:)`. Every error path therefore raised `ArgumentError` and returned **500** instead of the intended 4xx — this is the real root cause behind the AC4 inline SKU-uniqueness error never reaching the form (the symptom looked like a generic toast either way).

- Fix all `error_response` calls in `products_controller` and `products/variants_controller` to positional `code`/`message`.
- `validation_error_response` now emits `format_validation_errors` — the structured `[{ field, messages, full_messages }]` shape already used across the CRM (labels, message_templates, inboxes, pipeline_items, and the global `RecordInvalid` rescue), so the frontend can key errors by field.
- Sibling latent 500s fixed in the same files: product/variant **not-found** now returns 404, product/variant **in-use** on delete returns 422.

## Security
- No auth/permission changes. Endpoints keep their existing `require_permissions` gates.
- No new inputs; only error-response formatting.

## Test plan
- `crm: ruby -c` on both controllers — Syntax OK
- `crm: bundle exec rspec spec/requests/api/v1/products_validation_spec.rb` — 3 examples, 0 failures (422 duplicate SKU with structured details, 422 blank name, 404 unknown id)
- `crm: bundle exec rspec spec/requests/api/v1/products_bulk_spec.rb spec/requests/api/v1/products_validation_spec.rb` — 35 examples, 0 failures
- `crm: bundle exec rubocop` on both files — only pre-existing offenses (ClassLength/MethodLength/permissions-hash alignment), none on changed lines

## Changed Files
- `app/controllers/api/v1/products_controller.rb`
- `app/controllers/api/v1/products/variants_controller.rb`
- `spec/requests/api/v1/products_validation_spec.rb` (new)

## Related PRs
- frontend: https://github.com/evolution-foundation/evo-ai-frontend-community/pull/174

## Linked Issue
- EVO-1783

### DIFF
--- a/app/controllers/api/v1/products/variants_controller.rb
+++ b/app/controllers/api/v1/products/variants_controller.rb
@@ -50,9 +50,9 @@ class Api::V1::Products::VariantsController < Api::V1::BaseController
       )
     else
       error_response(
-        code: ApiErrorCodes::VALIDATION_ERROR,
-        message: 'Variant is in use and cannot be deleted',
-        details: @variant.errors.full_messages,
+        ApiErrorCodes::VALIDATION_ERROR,
+        'Variant is in use and cannot be deleted',
+        details: format_validation_errors(@variant.errors),
         status: :unprocessable_entity
       )
     end
@@ -64,8 +64,8 @@ class Api::V1::Products::VariantsController < Api::V1::BaseController
     @product = Product.find(params[:product_id])
   rescue ActiveRecord::RecordNotFound
     error_response(
-      code: ApiErrorCodes::RESOURCE_NOT_FOUND,
-      message: "Product with id #{params[:product_id]} not found",
+      ApiErrorCodes::RESOURCE_NOT_FOUND,
+      "Product with id #{params[:product_id]} not found",
       status: :not_found
     )
   end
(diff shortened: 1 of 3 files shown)


dpaes (approved):
## Re-review — APPROVED ✅ (round 2)

Paired with frontend#174. This fixes the real root cause behind EVO-1783 AC4: the products + variants controllers called `error_response` with keyword args (`code:`, `message:`) against the positional signature `error_response(code, message, details:, status:)` — so every error path raised `ArgumentError` → **500** instead of the intended 4xx.

**Verified:**
- All in-scope `error_response` calls switched to positional; `validation_error_response` now emits `format_validation_errors` → `[{field, messages, full_messages}]` under `error.details`, which the frontend keys by field.
- Bonus sibling fixes in the same files: not-found → 404, in-use-on-delete → 422.
- New request spec pins 422 (duplicate SKU + blank name) and 404 (unknown id), guarding the 500 regression. The duplicate-SKU case relies on `Product validates :sku, uniqueness: true, allow_blank: true` — confirmed.

**Non-blocking / out of scope (follow-up card filed):** the identical keyword-arg defect exists in ~22 other controllers (e.g. `conversations`, `agent_bots`, `automation_rules`, `webhooks`, `teams`, `macros`…), including two product-adjacent ones not touched here — `pipeline_items/products_controller.rb` and `ai_agents/products_controller.rb`. Each returns 500 instead of 4xx on error paths. Tracked separately.

**CI caveat:** crm-community does not run RSpec in CI — "35 examples, 0 failures" is self-reported; the new spec would catch the regression if run.

Approving and merging (squash).

PR #81, opened 2026-05-18
fix(crm): render-side label-UUID guard in activity messages (EVO-1190)

nickoliveira23:
## Summary

Add a render-time guard in `LabelActivityMessageHandler` so the conversation activity message never shows a UUID in place of a `Label#title`, regardless of how the label landed in `label_list`.

EVO-1190 was filed as a regression of EVO-1001, but the diagnosis (orphan UUID-named tags + `Automation System added <uuid>` activity messages dated 2026-05-06 → 2026-05-11) shows the bug was actually introduced by EVO-989 (per-stage automation rules, 6a3a926, 2026-05-05) and the primary write path was already fixed by 87d700a on 2026-05-12. This PR adds defense-in-depth for the **remaining** write paths that still skip the upstream resolvers:

- `BulkActionsJob#bulk_add_labels` and `#remove_labels` — bulk-action payload carries UUIDs straight to `Labelable`.
- `ActionService#remove_label` — operates on the existing `label_list`, so any legacy UUID stays in the diff.
- Direct writes in `app/controllers/api/v1/contacts_controller.rb` (lines 441/454).
- Legacy data persisted before EVO-1001 (orphan UUID-shaped `tags.name` rows observed in dev DB).

## Behaviour

- UUID-shaped entries are resolved to `Label#title` at activity-message compose time.
- Plain titles pass through unchanged.
- A UUID for a deleted label falls back to the raw value so the message still renders.

## Validation

- `evo-ai-crm-community: ruby -c app/models/concerns/label_activity_message_handler.rb` → Syntax OK
- `evo-ai-crm-community: bundle exec rspec spec/models/concerns/label_activity_message_handler_spec.rb spec/models/conversation_label_dispatch_spec.rb` → 10 examples, 0 failures (includes 2 E2E examples that exercise the bypass path via `conv.update!(label_list: [uuid])`)
- `evo-ai-crm-community: bundle exec rubocop app/models/concerns/label_activity_message_handler.rb spec/models/concerns/label_activity_message_handler_spec.rb` → 0 offenses

## Changed Files

- `app/models/concerns/label_activity_message_handler.rb`
- `spec/models/concerns/label_activity_message_handler_spec.rb`

## Follow-up (out of scope here)

[...]

PR #161, opened 2026-06-23
fix(contacts): backfill custom_attributes display-name keys to attribute_key (EVO-1850)

daniloleonecarneiro:
## Summary
Data migration that heals contacts whose `custom_attributes` were written under `attribute_display_name` by the journey Update node before EVO-1850, remapping the keys to the canonical `attribute_key` slug (for the `contact_attribute` model).

Guards:
- Skips ambiguous display names — a name mapping to >1 distinct key, or a display name colliding with another def's key (`attribute_display_name` has no DB uniqueness; only `(attribute_key, attribute_model)` is unique). Never guesses.
- **Collision-safe:** never clobbers an existing slug value (mixed write paths).
- Uses `update_column` to bypass the Wisper `custom_attributes` diff publisher, avoiding an event storm on bulk backfill.
- `down` is `IrreversibleMigration` (original display-name keys aren't recoverable once remapped).

## Test plan
- New `spec/migrations/...backfill_contact_custom_attribute_keys_spec.rb` — heal, preserve-other-attributes, collision-safe, ambiguity-safe, idempotency, irreversible-down. **6/6 green** (`RAILS_ENV=test rspec`).
- Caught + fixed a real bug via the spec: `contacts.type` is a `contact_type_enum` (not Rails STI), so the migration's inner `Contact` model needs `self.inheritance_column = :_type_disabled` or `find_each`/`create` blow up on "person".

## Notas
- Part of EVO-1850 (FE + evo-flow + this backfill).
- `schema.rb` change is only the migration version bump.

## Summary by Sourcery

Add a data migration to backfill contact custom attribute keys from display names to canonical attribute_key slugs with safety guards and an irreversible rollback.

Bug Fixes:
- Correct contacts whose custom_attributes were stored under attribute_display_name so they align with the canonical attribute_key-based access paths without overwriting existing slug values.

Enhancements:
- Introduce a guarded backfill that skips ambiguous or colliding display names and bypasses event publishing to avoid emitting unnecessary custom attribute change events.

Build:
- Bump schema.rb to the new migration version.

Tests:
- Add migration specs covering successful backfill, preservation of existing attributes, collision and ambiguity handling, idempotency, and irreversible down behavior.

### DIFF
--- a/db/migrate/20260623110717_backfill_contact_custom_attribute_keys.rb
+++ b/db/migrate/20260623110717_backfill_contact_custom_attribute_keys.rb
@@ -1,0 +1,116 @@
+# Backfill contact custom_attributes keyed by attribute_display_name (written by
+# the journey "Update Custom Attribute" node before EVO-1850) so they are keyed
+# by the canonical attribute_key slug — matching the CRM contact UI/API/import
+# and the journey Conditional read path ({{contact.customAttributes.<slug>}}).
+class BackfillContactCustomAttributeKeys < ActiveRecord::Migration[7.1]
+  # custom_attribute_definitions.attribute_model enum → contact_attribute.
+  CONTACT_ATTRIBUTE_MODEL = 1
+
+  class CustomAttributeDefinition < ActiveRecord::Base
+    self.table_name = 'custom_attribute_definitions'
+  end
+
+  class Contact < ActiveRecord::Base
+    self.table_name = 'contacts'
+    # `contacts.type` is a contact_type_enum column, NOT Rails STI — disable the
+    # inheritance lookup so find_each/create don't choke on "person"/"company".
+    self.inheritance_column = :_type_disabled
+  end
+
+  def up
+    unless table_exists?(:custom_attribute_definitions) && table_exists?(:contacts)
+      say 'Required tables missing; skipping backfill', true
+      return
+    end
+
+    remap = build_remap
+    if remap.empty?
+      say 'No display-name → attribute_key remappings needed; skipping', true
+      return
+    end
+
+    say "Remapping #{remap.size} display-name key(s) to attribute_key slug(s)", true
+    moved = 0
+    skipped = 0
+    contacts_touched = 0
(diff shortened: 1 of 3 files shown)


dpaes:
EVO-1850 review: this backfill migration is solid — correct enum (`contact_attribute: 1` confirmed in the model), `inheritance_column` disabled (the `type` column is an enum, not STI), `update_column` to bypass the Wisper diff publisher, and specs covering ambiguity/collision/idempotency/irreversibility. ✅

Not merging this round only because the card is held by a **data-loss blocker in evo-flow-community #81**: the journey Update node's read-modify-write reads `contact.customAttributes` off the raw snake_case `ContactDto` (real field `custom_attributes`), so it still replaces the whole `custom_attributes` column and wipes siblings on each run — which would re-corrupt exactly what this migration heals. Once #81 reads the real field, this migration can ride along. Details on the card (EVO-1850) and on #81.

Minor (non-blocking): `Contact.where.not(custom_attributes: nil)` will full-scan if the column defaults to `{}` — fine for a one-shot backfill, just noting it.

dpaes (approved):
## ✅ Approved (round 2)

Backfill migration is solid: `attribute_model` enum = `contact_attribute` (1), `inheritance_column` disabled so the `type` enum column doesn't trigger STI, ambiguity-safe (skips display names mapping to >1 key or colliding with another `attribute_key`), collision-safe (never clobbers an existing slug value), `update_column` to avoid the Wisper diff-event storm, idempotent (re-run is a no-op), irreversible `down` (documented). Specs cover ambiguity / collision / idempotence / irreversibility.

Heals existing contact data written under display-name keys → **AC4 (data) met.**

Scope note (non-blocking, N1): this repairs the stored **data**, not legacy journey **definitions** that still carry a display-name `attributeName`. Those re-write the display-name key on each run until re-saved (with attribute re-selection) or the runtime slug-resolution follow-up lands. Accepted-gap / follow-up.

PR #162, opened 2026-06-23
feat(crm): canned response attachments — read, remove, size limit (EVO-1861)

nickoliveira23:
## Summary
Canned response attachments were write-only: the serializer never returned them, so uploaded files were invisible on every read. This wires the read + remove paths and adds a server-side size guard.

- Serializer returns `attachments: attachments.map(&:push_event_data).compact` for index/show/create/update (matches the FE `CannedResponseAttachment` shape).
- Index eager-loads `attachments: { file_attachment: :blob }` to avoid N+1 when serializing.
- Detach path: `remove_attachment_ids[]` on update destroys the listed attachments, scoped to the record's own attachments (cannot touch another record's files); reloads the association so the response reflects the deletion.
- Server-side size limit: reject attachments > 10 MB (multipart + signed_id) — the modal cap was client-only and bypassable.
- Drive-by fix: `create`/`update` `error_response` passed `code:`/`message:` as keywords to a positional method, so every validation error 500'd instead of 422. Now positional.

Scope note: one attachment per canned response is enforced on the frontend (channels send a single media today); the backend intentionally keeps its capacity flexible (see EVO-1861 description).

## Security
- Detach is scoped to `@canned_response.attachments` — cannot delete another record's attachments.
- Server-side size limit closes a client-only-validation bypass.
- Permission gating unchanged (`canned_responses.update`); no new endpoints.

## Test plan
- `bundle exec rspec spec/requests/api/v1/canned_responses_attachments_spec.rb spec/requests/api/v1/canned_responses_show_spec.rb` — 6 examples, 0 failures (empty/populated serialization, detach, cross-record isolation, oversized → 422)
- `ruby -c` + `rubocop` on changed files

## Changed Files
- `app/serializers/canned_response_serializer.rb`
- `app/controllers/api/v1/canned_responses_controller.rb`
- `spec/requests/api/v1/canned_responses_attachments_spec.rb`

## Related PRs
- frontend: paired PR on `evo-ai-frontend-community` (see EVO-1861)

## Linked Issue
- EVO-1861

https://claude.ai/code/session_01Dk9VSPAVSDo3cG8caUbNND

## Summary by Sourcery

Expose and manage canned response attachments in the API while enforcing a server-side size limit and fixing validation error handling.

New Features:
- Return canned response attachments in the serializer payload for index, show, create, and update responses.
- Allow clients to remove canned response attachments via a remove_attachment_ids parameter on update.

Bug Fixes:
- Fix canned response create and update error handling so validation failures return a 422 response instead of a 500.

Enhancements:
- Eager-load canned response attachments and blobs in index requests to avoid N+1 queries when serializing.
- Normalize attachment parameter handling and add server-side rejection for attachments larger than 10 MB.

Tests:
- Add request specs covering canned response attachment serialization, detachment behavior, cross-record isolation, and oversized attachment validation.

### DIFF
--- a/app/controllers/api/v1/canned_responses_controller.rb
+++ b/app/controllers/api/v1/canned_responses_controller.rb
@@ -2,14 +2,16 @@ class Api::V1::CannedResponsesController < Api::V1::BaseController
   include FileTypeHelper
 
   require_permissions({
-    index: 'canned_responses.read',
-    show: 'canned_responses.read',
-    create: 'canned_responses.create',
-    update: 'canned_responses.update',
-    destroy: 'canned_responses.delete'
-  })
+                        index: 'canned_responses.read',
+                        show: 'canned_responses.read',
+                        create: 'canned_responses.create',
+                        update: 'canned_responses.update',
+                        destroy: 'canned_responses.delete'
+                      })
 
   before_action :fetch_canned_response, only: [:show, :update, :destroy]
+
+  MAX_ATTACHMENT_BYTES = 10.megabytes
 
   def index
     @canned_responses = canned_responses
@@ -31,8 +33,10 @@ class Api::V1::CannedResponsesController < Api::V1::BaseController
   end
 
   def create
+    return if reject_oversized_attachments
+
     @canned_response = CannedResponse.new(canned_response_params)
-    
+
     if @canned_response.save
       attach_files if params[:attachments].present?
       success_response(
@@ -42,8 +46,8 @@ class Api::V1::CannedResponsesController < Api::V1::BaseController
       )
     else
       error_response(
-        code: ApiErrorCodes::VALIDATION_ERROR,
-        message: 'Validation failed',
+        ApiErrorCodes::VALIDATION_ERROR,
+        'Validation failed',
         details: @canned_response.errors.full_messages,
         status: :unprocessable_entity
       )
@@ -51,6 +55,8 @@ class Api::V1::CannedResponsesController < Api::V1::BaseController
   end
 
   def update
+    return if reject_oversized_attachments
+
     if @canned_response.update(canned_response_params)
       detach_files if params[:remove_attachment_ids].present?
       attach_files if params[:attachments].present?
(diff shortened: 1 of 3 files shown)


dpaes:
**Review for EVO-1861 (backend)** — card moved to Todo (rejected at the card level due to the frontend AC4 gap); this PR stays open.

Backend looks good:
- Serializer returns attachments on index/show/create/update — explicit metadata (`id/file_type/data_url/...`), no `.attributes` over-exposure.
- `remove_attachment_ids[]` is association-scoped (no IDOR).
- 10 MB gate runs before persistence on both multipart and signed_id paths.
- Positional `error_response` fix is correct (validation errors now 422, not 500).
- Index eager-loads `attachments: { file_attachment: :blob }` (no N+1).

On the "server-side single-attachment cap": re-reading the AC ("limited to a single attachment; oversized files are rejected (client + server)"), the `(client + server)` qualifier applies to the **size** limit only — the 1-attachment cap is FE-only per the rescope. So this is in scope; not a blocker.

One medium worth fixing (ideally in the same round as the FE fix): an invalid/expired `signed_id` returns nil from `find_signed`, so `byte_size` is nil and it slips past the size gate, then `attach(nil)` raises → **500 instead of 422**. Guard the blob lookup and return a 422. (CI doesn't run RSpec here — please confirm the spec locally.)

dpaes (approved):
## Approved — round 2 ✅

Re-reviewed against the round-1 findings; the one backend item is resolved and verified against source.

**M3 (signed_id 500 → 422):** `reject_invalid_signed_ids` now validates `ActiveStorage::Blob.find_signed(...).nil?` and returns 422 before save/attach in both `create` and `update`; the request spec covers the invalid/expired case.

**Re-verified:**
- `error_response` positional fix confirmed against `api_response_helper.rb:82` (`def error_response(code, message, details:, status:)`) — validation errors now return 422, not 500.
- Detach is scoped to the record's own attachments (no IDOR; cross-record isolation spec passes).
- 10 MB server-side limit on both paths (multipart + signed_id).
- Index eager-loads `attachments: { file_attachment: :blob }` (no N+1).

**Non-blocking nit:** the signed_id is decoded three times on the happy path (reject_invalid → byte_size → attach). Negligible for a single attachment.

**Note:** specs are self-reported (this repo runs only Sourcery/contract checks in CI, no RSpec). Merge state CLEAN.

Approving and squash-merging.

PR #166, opened 2026-06-24
fix(crm): skip WhatsApp revoke/protocol messages on evolution_go inbound (EVO-1748)

nickoliveira23:
## Summary
When a WhatsApp contact deletes a message ("delete for everyone"), it arrives on the **evolution_go** inbound as a `Message` event carrying a `protocolMessage` (revoke). `message_processable?` passed because `@evolution_go_message` was present even though `message_content` was `nil`, so the revoke was created as an **empty-content message** ("the next message comes empty"). This skips protocol/control messages before message creation, so no empty bubble is created.

The `evolution` (v2) and `baileys` handlers already ignore protocol via `ignore_message?` — only evolution_go was affected, so the fix is scoped to it.

## Scope note (rescoped after smoke + product call with Daniel)
This card now covers ONLY the empty-bubble bug. Two distinct follow-ups were split out:
- Inbound revoke **visual feedback** (show "deleted by contact" while keeping the content) — separate UX card.
- Outbound delete **propagation** to the provider (delete-for-everyone) — separate card; only feasible on some providers (Baileys yes, evolution_go has no send-delete).

## Security
- No new inputs/endpoints; single-tenant.

## Test plan
- `evo-ai-crm-community: bundle exec rspec spec/services/whatsapp/message_revoke_handling_spec.rb` (3 examples, 0 failures)
- Manual smoke (evolution_go + live evolution-api): contact deletes a message → no empty bubble appears; following messages render normally.

## Changed Files
- `app/services/whatsapp/evolution_go_handlers/messages_upsert.rb`
- `spec/services/whatsapp/message_revoke_handling_spec.rb`

## Linked Issue
- EVO-1748

## Summary by Sourcery

Skip creation of CRM messages for WhatsApp protocol/control events on the evolution_go inbound handler to avoid empty bubbles when contacts delete messages.

Bug Fixes:
- Prevent empty WhatsApp message bubbles caused by processing evolution_go revoke/protocol messages as normal inbound messages.

Tests:
- Add specs covering WhatsApp message revoke handling to ensure protocol messages are skipped on evolution_go.

### DIFF
--- a/app/services/whatsapp/evolution_go_handlers/messages_upsert.rb
+++ b/app/services/whatsapp/evolution_go_handlers/messages_upsert.rb
@@ -8,6 +8,11 @@ module Whatsapp::EvolutionGoHandlers::MessagesUpsert
   private
 
   def handle_message
+    if protocol_message?
+      Rails.logger.info "Evolution Go API: Protocol message (control event, e.g. revoke) — skipping, not creating a message"
+      return
+    end
+
     return unless message_processable?
 
     Rails.logger.info "Evolution Go API: Creating new message #{raw_message_id}"
@@ -18,6 +23,10 @@ module Whatsapp::EvolutionGoHandlers::MessagesUpsert
     set_conversation
     update_conversation_status_if_needed
     create_message(attach_media: media_attachment?)
+  end
+
+  def protocol_message?
+    @evolution_go_message.is_a?(Hash) && @evolution_go_message[:protocolMessage].present?
   end
 
   def set_contact
--- a/spec/services/whatsapp/message_revoke_handling_spec.rb
+++ b/spec/services/whatsapp/message_revoke_handling_spec.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# EVO-1748: a WhatsApp revoke (delete-for-everyone) arrives via evolution_go as a
+# protocolMessage. It must NOT be processed as a regular message (which produced an
+# empty bubble). evolution v2 / baileys already skip protocol via ignore_message?.
+RSpec.describe 'WhatsApp evolution_go protocol message handling (EVO-1748)' do
+  let(:channel) { instance_double(Channel::Whatsapp, provider: 'evolution_go') }
+  let(:inbox) { instance_double(Inbox, id: 1, channel: channel) }
+  let(:service) { Whatsapp::IncomingMessageEvolutionGoService.new(inbox: inbox, params: { event: 'Message', data: {} }) }
+
+  before { service.instance_variable_set(:@inbox, inbox) }
+
+  describe '#protocol_message?' do
+    it 'is true for a revoke/protocol message' do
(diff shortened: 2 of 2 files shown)


dpaes (approved):
## Approved ✅

Minimal, correctly-targeted fix. No blockers. Resolves AC2 (empty bubble after a WhatsApp delete-for-everyone on `evolution_go`).

### Root cause — confirmed by reading the code
`message_processable?` (`helpers.rb:156`) returns `false unless message_content.present? || @evolution_go_message.present?`. On a revoke, `message_content` is `nil` but `@evolution_go_message.present?` is `true` (the `protocolMessage` hash is there) → it materialized as a `content: ''` message (the empty bubble). The new guard short-circuits before that. Correct.

### The key-form trap does NOT bite here (verified)
`protocol_message?` requires `@evolution_go_message.is_a?(Hash) && [:protocolMessage].present?`. Confirmed safe via three independent signals:
1. The same symbol/camelCase access idiom (`message[:protocolMessage]`, `message[:conversation]`) is already used and proven in `helpers.rb:351`/`:136` — text/media work in prod.
2. The service runs through `Webhooks::WhatsappEventsJob` (ActiveJob/Sidekiq); ActiveJob restores symbol keys (or yields `HashWithIndifferentAccess`) — both are `is_a?(Hash) == true`.
3. Your live smoke confirmed the bubble disappears → the guard fires in real runtime.

The `is_a?(Hash)` check is correct defensive coding: a `nil`/non-Hash falls through to the existing flow with no `NoMethodError`.

### Parity with evolution v2 / baileys — true, but partial (non-blocking)
`ignore_message?` in evolution v2 (`helpers.rb:144`) also skips `unsupported` and any blank-content non-media message; baileys (`helpers.rb:124`) also skips `context`/`edited`/blank `reaction`. This fix skips **only** `protocolMessage`. So a **reaction / unsupported / location / contacts** message on `evolution_go` still produces the same empty bubble (same symptom, different trigger). Out of scope for this card (revoke-specific); worth a future broadening to mirror evolution v2's `ignore_message?` if we want to close the whole class.

### Spec (non-blocking)
Pure unit test (sets `@evolution_go_message` directly). It doesn't exercise the real `process_evolution_go_message` → `data[:Message]` extraction or a full revoke payload through `perform`, and crm-community CI doesn't run RSpec, so regression cover is thin. Consider one integration example feeding `{ event: 'Message', data: { Info:{…}, Message: { protocolMessage:{…} } } }` through `perform` and asserting `inbox.messages` count is unchanged.

### Scope
Card AC1 ("deleted message reflected consistently") is intentionally deferred to **EVO-1890** (inbound "deleted by contact" notice) and **EVO-1891** (outbound delete propagation), both linked. This PR closes **AC2** only — confirmed acceptable. Note EVO-1890 will need the `protocolMessage` payload to identify the revoked message, so it will extend this guard rather than drop the event outright.

Merging (squash). Submodule pointer not bumped here — that goes in the batched release bump.

PR #167, opened 2026-06-24
feat(crm): WhatsApp contact-revoke notice + agent-delete propagation (EVO-1890, EVO-1891)

nickoliveira23:
Combined delivery for **EVO-1890** (inbound contact-revoke notice) + **EVO-1891** (outbound delete propagation). Pairs with the frontend PR (see Related PRs).

## Summary
- **EVO-1890 — inbound:** when a contact deletes a message ("delete for everyone"), the original message is **kept** and flagged `content_attributes.revoked_by_contact` (the frontend shows a "deleted by the contact" notice) instead of vanishing / leaving an empty bubble.
  - `evolution` + `baileys`: handled via the **`messages.delete`** webhook event (`process_messages_delete`). This is the real revoke path for Baileys-based gateways — confirmed live.
  - `evolution_go`: handled via the inline `protocolMessage` on the Message event.
  - Shared helper `mark_message_revoked_by_source_id` marks **only INCOMING** messages — guards against the provider echoing our own outbound delete back as a `fromMe` delete (which would mislabel the agent's own message).
- **EVO-1891 — outbound:** when an agent deletes an **outgoing** message, it is revoked on the provider (delete-for-everyone) where supported.
  - `Channel::Whatsapp#delete_message` → `EvolutionService#delete_message` → `DELETE /chat/deleteMessageForEveryone/{instance}`.
  - Runs in **`Whatsapp::DeleteMessageOnProviderJob`** (Sidekiq) so a slow/unreachable provider never blocks the CRM soft-delete. Records `revoke_propagated` for the UI.
  - `evolution_go` has no send-delete → stays CRM-only (`revoke_propagated=false`).

## Cross-channel feasibility (documented for the cards)
- ✅ WhatsApp via self-hosted gateways (evolution / baileys / evolution_go) — full WhatsApp client, surfaces revokes.
- ❌ WhatsApp Cloud (Meta official) — Meta does NOT webhook message deletions. Not possible.
- ❌ Telegram — Bot API does NOT notify the bot when a user deletes. Not possible.
- N/A email / SMS.

## Security
- No new untrusted inputs; revoke payload values are parameterized JSON to the provider. Single-tenant.

## Test plan
- `evo-ai-crm-community: bundle exec rspec spec/services/whatsapp/message_revoke_handling_spec.rb spec/services/whatsapp/providers/evolution_service_delete_spec.rb spec/jobs/whatsapp/delete_message_on_provider_job_spec.rb` (15 examples, 0 failures)
- Manual smoke (evolution provider, live evolution-api): contact deletes → message kept + "deleted by contact"; agent deletes own outgoing message → removed on WhatsApp; agent's own delete no longer mislabeled.

## Changed Files
- `app/services/whatsapp/incoming_message_base_service.rb`, `incoming_message_evolution_service.rb`
- `app/services/whatsapp/{evolution,baileys,evolution_go}_handlers/*`
- `app/services/whatsapp/providers/evolution_service.rb`
- `app/models/channel/whatsapp.rb`, `app/models/message.rb`
- `app/controllers/api/v1/conversations/messages_controller.rb`
- `app/jobs/whatsapp/delete_message_on_provider_job.rb`
- specs (revoke handling, evolution delete, job)

## Related PRs
- frontend: see EVO-1890

## Linked Issues
- EVO-1890, EVO-1891

## Summary by Sourcery

Handle WhatsApp message revocations end-to-end, keeping revoked inbound messages visible with a flag and propagating agent-initiated deletions to supported WhatsApp providers.

New Features:
- Mark inbound WhatsApp messages as revoked-by-contact while retaining their content so the UI can show a deletion notice instead of removing the message.
- Propagate agent-initiated deletions of outgoing WhatsApp messages to the Evolution provider as delete-for-everyone operations, recording whether revocation succeeded for UI display.

Enhancements:
- Share revoke-handling helpers across WhatsApp providers to extract revoked message IDs and mark only incoming messages as revoked.
- Extend Evolution, Baileys, and Evolution Go handlers to process protocol and messages.delete events as revocations instead of creating empty messages or ignoring them.

Tests:
- Expand WhatsApp revoke-handling specs to cover inbound revocation marking, protocol-message handling, and provider-specific delete events.
- Add specs for Evolution provider message deletion and the background job that propagates delete-for-everyone operations to WhatsApp.

### DIFF
--- a/app/controllers/api/v1/conversations/messages_controller.rb
+++ b/app/controllers/api/v1/conversations/messages_controller.rb
@@ -47,6 +47,7 @@ class Api::V1::Conversations::MessagesController < Api::V1::Conversations::BaseC
   def destroy
     @message = message
     @message.update!(content_attributes: (@message.content_attributes || {}).merge(deleted: true))
+    enqueue_provider_delete(@message)
 
     success_response(
       data: MessageSerializer.serialize(@message, include_attachments: true, include_sender: true),
@@ -91,6 +92,18 @@ class Api::V1::Conversations::MessagesController < Api::V1::Conversations::BaseC
   end
 
   private
+
+  # EVO-1891: when an agent deletes an OUTGOING message, revoke it on the provider
+  # (delete-for-everyone) where supported. Done in a background job so a slow or
+  # unreachable provider never blocks/delays the CRM soft-delete response.
+  def enqueue_provider_delete(message)
+    return unless message.outgoing?
+
+    channel = message.conversation&.inbox&.channel
+    return unless channel.respond_to?(:delete_message)
+
+    Whatsapp::DeleteMessageOnProviderJob.perform_later(message.id)
+  end
 
   def perform_status_update(target_status)
     Messages::StatusUpdateService.new(@message, target_status, permitted_params[:external_error]).perform
--- a/app/jobs/whatsapp/delete_message_on_provider_job.rb
+++ b/app/jobs/whatsapp/delete_message_on_provider_job.rb
@@ -1,0 +1,18 @@
+# EVO-1891: propagate an agent's outgoing message deletion to the WhatsApp
+# provider (delete-for-everyone) OUTSIDE the request cycle, so a slow or
+# unreachable provider never blocks the CRM soft-delete. Records whether the
+# revoke actually propagated so the UI can flag CRM-only deletions.
+class Whatsapp::DeleteMessageOnProviderJob < ApplicationJob
+  queue_as :low
+
+  def perform(message_id)
+    message = Message.find_by(id: message_id)
+    return unless message&.outgoing?
+
+    channel = message.conversation&.inbox&.channel
+    return unless channel.respond_to?(:delete_message)
+
+    propagated = channel.delete_message(message)
+    message.update!(content_attributes: message.content_attributes.merge(revoke_propagated: propagated))
+  end
+end
--- a/app/models/channel/whatsapp.rb
+++ b/app/models/channel/whatsapp.rb
@@ -85,6 +85,18 @@ class Channel::Whatsapp < ApplicationRecord
     end
   end
 
+  # Propagates an agent's message deletion to the provider (delete-for-everyone),
+  # where supported. Returns false (CRM-only) for providers without delete_message.
+  def delete_message(message)
+    service = provider_service
+    return false unless service.respond_to?(:delete_message)
+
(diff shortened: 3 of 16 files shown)


dpaes (changes requested):
## Changes requested — EVO-1890 AC1 (the notice must verifiably surface)

Rejecting on one thing only: EVO-1890's AC1 — that the agent actually **sees** the "deleted by contact" notice. The revoke-marking wiring is structurally sound (the incoming-only guard correctly avoids mislabeling the agent's own messages), but nothing proves the revoke key id matches the stored `source_id`: every spec stubs `inbox.messages.find_by(source_id:)`, and the card's own smoke note says the mark wasn't surfacing (likely an id-format mismatch).

**What's missing:** verify against a real captured revoke payload (per active provider) that `revoked_message_source_id` equals the stored `source_id`, and add a test that does **not** stub that lookup.

**Non-blocking (can wait — secondary provider):** EVO-1891 Z-API — `Channel::Whatsapp#delete_message` calls `service.delete_message(message)` (1 arg) but `ZapiService#delete_message(phone, id)` takes 2 → `ArgumentError` swallowed by the rescue → `revoke_propagated:false` (silent no-op, while Z-API is treated as "supported"). Fine to address in a follow-up.

dpaes (approved):
## Re-review — Approved (AC1 addressed)

The AC1 reject (the "deleted by contact" notice not proven to surface) is resolved. New spec `spec/services/whatsapp/evolution_revoke_integration_spec.rb` proves the flow end-to-end **without stubbing `find_by`**:
- creates a real `Message` with `source_id`, runs the real `messages.delete` event through `IncomingMessageEvolutionService`, reloads and asserts `revoked_by_contact == true` + content preserved;
- **test 3 (decisive):** ingests via the real `messages.upsert`, then revokes by the **same key id**, proving `inbound source_id == delete id` — the exact id-format concern from the earlier smoke;
- incoming-only guard (fromMe) + no-match no-op, both with real persistence.

Caveat: couldn't run it locally this session (Docker engine offline), so this rests on the test design + author's run + the code path I read (`revoked_message_source_id` reads `key.id`; inbound stores `source_id = key.id`). CRM CI runs no rspec.

Non-blocking follow-ups (as previously agreed): the product flag (agent-deleted content show/hide, fe#192) and the Z-API `delete_message` arity (EVO-1891).

PR #176, opened 2026-06-25
fix(labels): harden Labelable write paths to always persist taggings (EVO-1932)

daniloleonecarneiro:
## Summary

`Contact#update_labels`/`#add_labels` live in `Labelable` (shared by **Contact, Conversation, Product**). They could return success while persisting **no tagging** (`save` => true, `taggings.count == 0`) — the journey add-label/remove-label false-success EVO-1932 isolated.

Root cause (model layer):
- The `acts_as_taggable_on` `label_list=` setter only writes taggings for inputs it can coerce to tag names. Any malformed token routed straight into the setter — `nil`, blank/whitespace strings, a `Tag` record, a symbol — became a silent no-op while `save` still returned true.
- `add_labels` made it worse: it unioned against `labels` (the **Tag association**, returning `ActsAsTaggableOn::Tag` records) instead of `label_list` (the string TagList), feeding `Tag` objects into the setter.

Note: EVO-1928 fixed this at the **controller** seam (`LabelConcern#create` normalising singular/scalar shapes so `update_labels(nil)` no longer wiped the list). EVO-1932 hardens the **model** so the contract holds for every caller (journeys, bulk actions, products, imports), not just the HTTP path.

## What changed
- `update_labels` / `add_labels` now normalise input to an array of non-blank strings before assigning.
- `add_labels` builds on `label_list` (strings), not the `labels` Tag association; union is deduped/idempotent.
- Added a symmetric `remove_labels`.
- Kept the **dirty-tracking setter path** (`update!(label_list: ...)`) rather than in-place `label_list.add`, so Conversation/Contact callbacks keyed off `saved_change_to_label_list?` (cached label list, label activity messages, EvoFlow `label.added/removed`) keep firing. In-place mutation would persist the tagging but NOT dirty-track, regressing those side effects.

## Test plan
New `spec/models/concerns/labelable_spec.rb` (11 examples, all green) asserts persistence through the public API:
- `update_labels(["teste label"])` on a fresh contact → reload `taggings.count >= 1` (primary AC); idempotency; REPLACE; `[]` clears; blank/whitespace normalised away.
- `add_labels` persists, unions without dropping existing, idempotent, accepts a bare scalar.
- `remove_labels` removes + persists smaller set; no-op for absent label.

```
$ bundle exec rspec spec/models/concerns/labelable_spec.rb
11 examples, 0 failures
```

rails runner evidence (patched model, new contact, before/after taggings):
```
NEW contact taggings(before)=0
after update_labels(['teste label']): reload taggings=1 label_list=["teste label"]
after add_labels(['extra']):          taggings=2 label_list=["teste label", "extra"]
after remove_labels(['teste label']): taggings=1 label_list=["extra"]
```

## Notes
- `Labelable` is shared by Conversation/Product; the normalise-then-assign approach preserves their existing setter-path semantics and dirty-tracking.
- The pre-existing `spec/controllers/concerns/label_concern_spec.rb:64` failure on develop is unrelated (an over-strict `array_including` matcher in a UUID-merge case) and untouched by this PR — EVO-1932 is model-only.

### DIFF
--- a/app/models/concerns/labelable.rb
+++ b/app/models/concerns/labelable.rb
@@ -9,13 +9,47 @@ module Labelable
   # (see `Contact#publish_label_changes`). Every write path that mutates
   # `label_list` and persists hits that callback, so update_labels/add_labels
   # no longer need to emit explicitly.
+  #
+  # EVO-1932: persistence robustness. `acts_as_taggable_on` only writes
+  # taggings when the `label_list` *setter* runs and dirty-tracks the virtual
+  # attribute — `update!(label_list: array)`/`record.label_list = array`. We
+  # deliberately keep that setter path (rather than in-place `label_list.add`)
+  # because Conversation/Contact callbacks key off `saved_change_to_label_list?`
+  # (cached label list, label activity messages, EvoFlow `label.added/removed`
+  # events); in-place mutation would persist the tagging but NOT dirty-track,
+  # silently dropping those side effects. To stop the setter receiving values
+  # the gem can't turn into a tagging — `nil`, blanks, `Tag` records, symbols —
+  # every entry point normalises to an array of non-blank strings first. The
+  # journey add-label/remove-label node reaches `update_labels` by NAME, so a
+  # malformed token must never become a false-success that fails to persist.
+
+  # REPLACE the contact/conversation/product label set with `labels`.
+  # `[]` (or all-blank input) clears the set — this is how the UI removes the
+  # last label and how a re-post of the desired set deletes a label.
   def update_labels(labels = nil)
-    update!(label_list: labels)
+    update!(label_list: normalize_label_tokens(labels))
   end
 
+  # ADD `new_labels` to the existing set (union, idempotent). Builds on
+  # `label_list` (the string TagList), NOT the `labels` Tag association, so the
+  # setter receives plain strings rather than `Tag` records.
   def add_labels(new_labels = nil)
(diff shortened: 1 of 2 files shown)


dpaes (approved):
## Approved

Verified at PR state: **model-only** (`labelable.rb` + new spec), no controller change → no conflict with #175/EVO-1928.
- Keeps the setter path `update!(label_list:)` (deliberate, lines 15-16) — correct: it preserves the dirty-tracking that fires `cached_label_list` / label activity / EvoFlow `label.added|removed`; in-place `label_list.add` would persist the tagging but skip those side-effects.
- Normalizes tokens to non-empty strings (`Array(tokens).map(&:to_s).reject(&:blank?)`), fixes `add_labels` to union on `label_list` (was pushing `Tag` objects via the `labels` association), adds symmetric `remove_labels`.
- 11 model specs in `evolution_test`.

Note on the card premise: "`update!(label_list:)` is a no-op" did **not** reproduce on develop (confirmed by the author across 5 rails-runner scenarios). The real false-success was the controller sending `update_labels(nil)` on the singular `labelId` shape (fixed in #175/EVO-1928) plus a **Puma-dev-only artifact** where the `acts_as_taggable_on` `save_tags` after_save callback is inactive under `eager_load=false`; it persists in-process, in specs, and in production. So EVO-1897/#169 persistence holds in prod, and this PR is a sound hardening on top.

PR #187, opened 2026-07-01
fix(crm): conversations#show N+1 — trim full thread (EVO-1972)

nickoliveira23:
## Summary

`conversations#show` was serializing the ENTIRE message thread (`include_messages: true`) with each message's sender/attachments not preloaded — a ~6s N+1 across ActiveStorage on every open. The frontend already loads messages via the paginated messages route, so the inline thread on `show` is dead weight.

- `show` → `include_messages: false`.
- `create` KEEPS `include_messages: true` (evo-flow / campaigns read `messages[0].id`).

Measured 6s → 342ms in-browser.

## Validation

- `ruby -c app/controllers/api/v1/conversations_controller.rb` — Syntax OK
- Adversarial code review confirmed nothing consumes `show`'s serialized `messages`, and `create` still returns `messages[0]`.

## Scope note

The LIST/index serializer N+1 (`agent_bot_active`) is EVO-1958 (Marcelo) — a different endpoint and root cause; no collision (disjoint file regions).

## Linked Issues

- EVO-1972

## Summary by Sourcery

Enhancements:
- Reduce conversations show response payload by disabling message inclusion, relying on the existing paginated messages endpoint instead.

### DIFF
--- a/app/controllers/api/v1/conversations_controller.rb
+++ b/app/controllers/api/v1/conversations_controller.rb
@@ -172,7 +172,12 @@ class Api::V1::ConversationsController < Api::V1::BaseController
     success_response(
       data: ConversationSerializer.serialize(
         @conversation,
-        include_messages: true,
+        # include_messages: false — o front carrega mensagens pela rota paginada
+        # /conversations/:id/messages; serializar a thread inteira aqui (e sem
+        # preload de :messages) causava N+1 de ~6s por request. Nenhum consumidor
+        # de show lê .messages (verificado no sistema inteiro). O CREATE mantém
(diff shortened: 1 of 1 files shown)


dpaes (approved):
Approved — one-line, high-confidence fix.

`conversations#show` → `include_messages: false` (create keeps `true` for the evo-flow/campaigns `messages[0].id` contract). Independently verified no consumer reads show's serialized messages: frontend `getConversation` callers ignore `.messages` (the thread loads via the paginated /messages route), evo-flow `conditional.node` reads only `.pipelines`, and the processor uses /labels. No regression; removes the ~6s N+1 across ActiveStorage on open.

PR #242, opened 2026-07-22
fix(crm): persist pipeline is_active and stop reporting false success (EVO-2122)

nickoliveira23:
## Summary

`PATCH /pipelines/:id` never persisted `is_active`: the attribute was missing from the strong-params permit, so Rails dropped it silently, `update!({})` became a no-op and the endpoint still answered `200` with `success: true`. The pipeline stayed active while the UI reported success.

- `:is_active` is now permitted on `update`. It is deliberately **not** permitted on `create` — a pipeline is always born active.
- An update whose payload reduces to zero permitted attributes is rejected with `422 INVALID_PARAMETER` instead of reporting success for something that never happened.
- `index` accepts `include_inactive=true`. The default still hides inactive pipelines, so every existing picker (dashboard, kanban, automations, agents, CRM forms) is untouched — only the pipelines management screen opts in, which is what keeps a deactivated pipeline reachable for reactivation.
- `pipeline_params` is memoized, since it is now read twice per update request and re-ran the whole `custom_fields` normalization.

The pre-existing `PATCH /pipelines/:id/archive` route is left in place and still works; it is covered by a regression spec.

## Validation

- `evo-ai-crm-community: bundle exec rspec spec/requests/api/v1/pipelines_activation_spec.rb` — 12 examples, 0 failures
- `evo-ai-crm-community: bundle exec rubocop app/controllers/api/v1/pipelines_controller.rb` — 41 offenses vs 42 on the baseline commit, no new cop triggered
- `evo-ai-crm-community: ruby -c app/controllers/api/v1/pipelines_controller.rb` — Syntax OK

## Changed Files

- `app/controllers/api/v1/pipelines_controller.rb`
- `spec/requests/api/v1/pipelines_activation_spec.rb` (new)

## Out of scope

Gating automations by `is_active` is a separate product decision, already parked in a comment on the issue. This PR only makes the flag persist.

## Linked Issue

- EVO-2122 — https://linear.app/evoai/issue/EVO-2122

## Summary by Sourcery

Ensure pipeline activation status is properly persisted and exposed while preventing no-op updates from reporting success.

Bug Fixes:
- Persist the is_active flag when updating a pipeline so deactivation and reactivation actually take effect.
- Return a 422 INVALID_PARAMETER error when a pipeline update payload contains no permitted attributes instead of falsely reporting success.

Enhancements:
- Allow the pipelines index endpoint to optionally include inactive pipelines via an include_inactive flag while keeping them hidden by default.
- Memoize pipeline_params and extend permitted attributes to support is_active updates without affecting creation behavior.

Tests:
- Add request specs covering pipeline activation/deactivation, serialization of is_active, protection against no-op updates, legacy archive behavior, and index filtering of inactive pipelines.

### DIFF
--- a/.github/workflows/spec-staleness-guard.yml
+++ b/.github/workflows/spec-staleness-guard.yml
@@ -30,6 +30,11 @@ on:
       - 'app/controllers/api/v1/contacts_controller.rb'
       - 'app/jobs/delete_object_job.rb'
       - 'spec/jobs/delete_object_job_spec.rb'
+      # EVO-2122: dropping is_active from the update permit made deactivation a
+      # silent no-op that still answered 200. The regression is invisible to every
+      # other lane, so the controller has to trigger the spec that proves it.
+      - 'app/controllers/api/v1/pipelines_controller.rb'
+      - 'spec/requests/api/v1/pipelines_activation_spec.rb'
 
 permissions:
   contents: read
@@ -98,6 +103,7 @@ jobs:
             spec/models/pipeline_spec.rb \
             spec/models/role_spec.rb \
             spec/requests/api/v1/contacts_spec.rb \
+            spec/requests/api/v1/pipelines_activation_spec.rb \
             spec/jobs/delete_object_job_spec.rb \
             spec/channels/room_channel_spec.rb \
             spec/lib/global_config_service_spec.rb \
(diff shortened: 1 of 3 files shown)


nickoliveira23:
## Related PRs

- Frontend: https://github.com/evolution-foundation/evo-ai-frontend-community/pull/263 — depends on the `include_inactive` param added here, so this one merges first.

gomessguii (approved):

PR #243, opened 2026-07-22
feat(crm): make pipeline archiving informed and preserve capture-form attribution (EVO-2200)

nickoliveira23:
## Summary

Card was re-scoped mid-flight — see the decision comment on EVO-2200. The original proposal (reject capture into an archived pipeline) was declined: rejecting discards the visitor's data, and skipping the pipeline item would break the form's Leads tab, which does not filter `is_active` and shows the lead today. **Runtime capture behaviour is unchanged.** What was missing is prevention.

- `GET /pipelines/:id/dependents` answers what still feeds a pipeline, so the archive confirmation can name it. The payload declares which dependency kinds were inspected — automations and journeys are separate children of EVO-2199, and a dialog implying it checked them would be worse than no dialog.
- The lookup covers both the form's default destination **and** routing rules that override it (`CrmForm#resolve_destination`). Matching only `default_pipeline_id` would report an all-clear while rule-routed leads kept arriving. Uses the existing GIN index on `routing_rules`.
- Capture into an archived pipeline is logged, so the proceed is observable.
- The originating form is mirrored on the contact as `capture_form_slugs` (a list — one person may fill several forms).

### Why the contact stamp matters

`form_slug` lived only in the pipeline item's `custom_fields`, and `DELETE /api/v1/pipeline_items/:id` is a hard delete with no soft-delete on the model. Deleting a kanban card destroyed the form attribution permanently — on a healthy, active pipeline. This stops that from the point forward, and is the prerequisite for the contact-based read (separate card).

## Security

- `dependents` is gated on `pipelines.update`, the operation it serves — not on read.
- Form names belong to the `crm_forms` resource, so they are withheld when the caller lacks `crm_forms.read`; only counts are shared (`names_redacted`).
- No new user input reaches a query; the JSONB match is parameterised.
- The list is capped at 50; counts remain exact.

## Test plan

- `bundle exec rspec spec/services/public spec/requests/public spec/requests/api/v1/pipeline_dependents_spec.rb spec/requests/api/v1/pipelines_activation_spec.rb spec/models/crm_form_spec.rb` — 64 examples, 0 failures
- `bundle exec rubocop app/services/public/leads/creation_service.rb app/controllers/api/v1/pipelines_controller.rb` — 53 offenses, identical to the baseline on `develop`; no new cop
- `ruby -c` on every touched file — Syntax OK
- Manual smoke against a live install: capture with the pipeline active and archived, deactivation dialog listing a real form, and attribution surviving deletion of the kanban card

## Deferred

`stamp_capture_form_on_contact` does a read-modify-write without a lock. Two simultaneous submissions by the same person through different forms could drop one slug. Serialising a contact on an anonymous public endpoint carries its own risk, and the scenario requires the same email hitting two forms at the same instant — recorded rather than fixed.

## Changed Files

- `app/controllers/api/v1/pipelines_controller.rb`
- `app/services/public/leads/creation_service.rb`
- `config/routes.rb`
- `spec/requests/api/v1/pipeline_dependents_spec.rb` (new)
- `spec/services/public/leads/creation_service_spec.rb`

## Related PRs

- Frontend: https://github.com/evolution-foundation/evo-ai-frontend-community/pull/264 — merge this one first, it depends on the `dependents` endpoint.

## Linked Issue

- EVO-2200 — https://linear.app/evoai/issue/EVO-2200

## Summary by Sourcery

Add backend support to surface CRM forms that depend on a pipeline before archiving it and to persist capture-form attribution on contacts while keeping lead capture into archived pipelines observable but unchanged.

New Features:
- Expose a pipelines dependents endpoint that reports CRM forms targeting a given pipeline, including counts and limited form metadata when permitted.
- Persist capture form slugs on contacts when leads are created so attribution survives pipeline item deletion.

Enhancements:
- Log captures into archived pipelines to make continued intake into inactive pipelines observable without blocking it.

Tests:
- Add request specs for the pipeline dependents endpoint and extend public leads creation service specs to cover contact form attribution, archived pipeline logging, and related behaviors.

### DIFF
--- a/app/controllers/api/v1/pipelines_controller.rb
+++ b/app/controllers/api/v1/pipelines_controller.rb
@@ -12,7 +12,7 @@ class Api::V1::PipelinesController < Api::V1::BaseController
     stats: 'pipelines.read',
     by_contact: 'pipelines.read',
     by_conversation: 'pipelines.read',
-    dependents: 'pipelines.read'
+    dependents: 'pipelines.update'
   })
 
   before_action :fetch_pipeline, only: [:show, :update, :destroy, :archive, :set_as_default, :dependents]
@@ -133,12 +133,15 @@ class Api::V1::PipelinesController < Api::V1::BaseController
   # covered today — automations and journeys are separate cards (EVO-2199), so the
   # payload names what it inspected instead of implying the list is exhaustive.
   def dependents
-    forms = CrmForm.where(default_pipeline_id: @pipeline.id).order(:name)
+    scope = forms_targeting_pipeline
 
     success_response(
       data: {
         inspected: ['crm_forms'],
-        crm_forms: forms.map { |f| { id: f.id, name: f.name, title: f.title, published: f.published } }
+        count: scope.count,
+        published_count: scope.where(published: true).count,
+        names_redacted: !may_read_forms?,
+        crm_forms: may_read_forms? ? serialize_dependent_forms(scope.limit(DEPENDENTS_LIMIT)) : []
       },
       message: 'Pipeline dependents retrieved successfully'
     )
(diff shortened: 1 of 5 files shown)


nickoliveira23:
## Related PRs

- Frontend: https://github.com/evolution-foundation/evo-ai-frontend-community/pull/264 — calls the `dependents` endpoint added here, so this PR merges first.

gomessguii (approved):

PR #319, opened 2026-08-28
fix(locale): use the inbox's configured locale for auto-sent template messages

mateusbellozupko:
## Summary
- `MessageTemplates::HookExecutionService` runs from an async event listener, outside any controller request — it never inherits the `I18n.locale` that `SwitchLocale` sets for HTTP requests. Every auto-sent template message (greeting, out-of-office, email-collect, CSAT survey) was rendered in `I18n.default_locale` (English), regardless of the inbox's configured locale (e.g. `pt_BR`), producing mixed-language conversations for non-English installs.
- Fix: resolve the inbox's `channel.locale` (falling back to the app default when unset/unsupported, mirroring `SwitchLocale#validate_and_get_locale`'s own fallback logic) and wrap all four template-generation calls in `I18n.with_locale(...)`.

## Testing notes
- Verified against a live pt_BR Web Widget inbox: before the fix, `email_collect`'s auto-sent messages ("Give the team a way to reach you." / "Get notified by email") rendered in English even though `pt.yml`/`pt_BR.yml` both already have correct translations for these exact keys — confirming the locale was simply never being set, not a missing translation.

## Summary by Sourcery

Render asynchronous template messages using each inbox's configured, validated locale.

Bug Fixes:
- Ensure automatically generated greeting, out-of-office, email-collection, and CSAT messages use the inbox's configured locale instead of defaulting to English.
- Fall back safely to the application locale when the inbox locale is unset or unsupported.

### DIFF
--- a/app/services/message_templates/hook_execution_service.rb
+++ b/app/services/message_templates/hook_execution_service.rb
@@ -14,10 +14,29 @@ class MessageTemplates::HookExecutionService
   delegate :contact, to: :conversation
 
   def trigger_templates
-    ::MessageTemplates::Template::OutOfOffice.new(conversation: conversation).perform if should_send_out_of_office_message?
-    ::MessageTemplates::Template::Greeting.new(conversation: conversation).perform if should_send_greeting?
-    ::MessageTemplates::Template::EmailCollect.new(conversation: conversation).perform if inbox.enable_email_collect && should_send_email_collect?
-    ::MessageTemplates::Template::CsatSurvey.new(conversation: conversation).perform if should_send_csat_survey?
+    # This runs from an async event listener, outside any controller request,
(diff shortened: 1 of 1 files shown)


mateusbellozupko:
Fixed in 77af7bf — `template_locale` now guards with `channel.respond_to?(:locale)` instead of bare safe-navigation, since most channel types don't define `#locale` at all.

mateusbellozupko:
@sourcery-ai review — the finding was already addressed on this branch: `template_locale` now guards with `channel.respond_to?(:locale)` before reading it, instead of relying on safe navigation alone (which doesn't help for a present channel that simply lacks the method, e.g. `Channel::Api`). Requesting a fresh review to clear the stale blocking status.

PR #320, opened 2026-08-28
fix(widget): fall back to the identified contact's own email/phone in pre-chat validation

mateusbellozupko:
## Summary
- `Api::V1::Widget::BaseController#contact_email` and `#contact_phone_number` only ever read from the incoming request's `contact` params — unlike `#contact_name`, they had no fallback to the already-identified `@contact`'s own stored data.
- The widget frontend's `PreChatForm` correctly omits pre-chat fields it already knows about a returning, identified contact (`has_email`/`has_phone_number`), so it doesn't resend them on a new conversation. With required pre-chat fields configured, this made `Widget::PreChatFormValidator` reject every new conversation from a known/returning contact with `PRE_CHAT_VALIDATION_ERROR` — the message appeared to send in the UI (optimistic state) but no conversation/message was ever actually created, and the bot never replied.
- Fix: both methods now fall back to the identified contact's existing `email`/`phone_number` when the request doesn't resend them, mirroring the existing `contact_name` pattern.

## Testing notes
- Reproduced directly against `POST /api/v1/widget/conversations` with a valid `X-Auth-Token` for an already-identified contact but no `contact` params: before the fix, this reliably returned `PRE_CHAT_VALIDATION_ERROR` ("Nome Completo is required", "Número de Telefone is required") even though the contact already had both on file.

## Summary by Sourcery

Restore widget conversation creation for returning contacts by falling back to their existing contact details during pre-chat validation.

Bug Fixes:
- Allow widget pre-chat validation to use an identified contact’s stored email, phone number, and name when those fields are omitted from the request.
- Prevent returning identified contacts from being blocked from creating conversations when required pre-chat fields are already known.

### DIFF
--- a/app/controllers/api/v1/widget/base_controller.rb
+++ b/app/controllers/api/v1/widget/base_controller.rb
@@ -88,7 +88,7 @@ class Api::V1::Widget::BaseController < ApplicationController
   end
 
   def contact_email
-    permitted_params.dig(:contact, :email)&.downcase || @contact.email.presence
+    permitted_params.dig(:contact, :email)&.downcase&.presence || @contact.email.presence
   end
 
   def contact_name
@@ -104,7 +104,7 @@ class Api::V1::Widget::BaseController < ApplicationController
     # so requiring them again here made pre-chat validation fail for every
     # returning contact starting a new conversation once pre-chat fields are
     # marked required (EVO-2233).
-    permitted_params.dig(:contact, :phone_number) || @contact.phone_number.presence
+    permitted_params.dig(:contact, :phone_number).presence || @contact.phone_number.presence
   end
 
   def browser_params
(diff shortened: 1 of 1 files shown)


mateusbellozupko:
Fixed in 945bd38 — both `contact_email` and `contact_phone_number` now call `.presence` on the incoming value before falling back to the identified contact's stored data, so an explicit empty string no longer suppresses the fallback.

mateusbellozupko:
Also fixed `contact_name` in 23a39b9 — same bug class, its early-return for already-identified contacts returned `nil` unconditionally instead of falling back to the contact's stored name, rejecting every returning contact when `fullName` is a required pre-chat field. Found this while investigating a live report where a returning contact with a saved name still got `"Nome Completo is required"`.

PR #379, opened 2026-09-19
fix(pipelines): stop misrouting pipeline_item moves to the wrong conversation

mateusbellozupko:
## Summary
This started as a card-ordering fix and turned up the actual root cause of intermittent "wrong card moves" / "card won't move" reports on the Kanban board.

**Root cause** (`app/controllers/api/v1/pipeline_items_controller.rb#set_pipeline_item`): the lookup for `move_to_stage`/`destroy`/`update_conversation` first tries `Conversation.find_by(display_id: params[:id])`. `display_id` is an integer column, and Rails casts a non-numeric string to its **leading digits** rather than `nil` — so a pipeline_item UUID that happens to start with digits (e.g. `"8a4ab330-...".to_i == 8`) silently matches whichever *unrelated* conversation has that display_id, and the move gets redirected to a completely different item. The item the caller actually specified is never touched. This reproduced 100% of the time for any pipeline_item whose UUID prefix collided with an existing conversation's display_id — confirmed directly against production data where every affected card matched this pattern exactly.

Fixed by only attempting the display_id lookup when the id param is actually a plain integer string.

**Also included**: `Pipeline#pipeline_items` / `PipelineStage#pipeline_items` now order by `(entered_at, id)` instead of relying on Postgres' unordered default, so the Kanban board renders cards in a stable, deterministic order across reloads (this was a real but secondary issue — an unordered association let cards reshuffle position between requests).

## Test plan
- [x] Added a controller spec reproducing the UUID/display_id collision exactly (fails without the fix, passes with it)
- [x] Added model specs proving `pipeline_items` ordering is stable, including the tied-`entered_at` case (bulk API-imported leads)
- [x] Ran the full pipeline model/controller/request spec suite — 102 examples, 0 failures (one pre-existing unrelated failure confirmed present on `main` without any of these changes)
- [x] Verified against live production data: every previously "stuck" card matched the exact digit-collision pattern this fix addresses


mateusbellozupko:
Good catch — pushed a follow-up commit adding `id` as a tiebreaker (`order(:entered_at, :id)`) plus a test covering the tied-`entered_at` case, since that's exactly what bulk API-imported leads hit.

